PR #29, opened 2024-11-07
Don't always crit (ranged and melee)

worm-girl:
The check for monsters to resist crits is now the higher of their melee * 5 or their ( dodge * mods * 5 ) 1.2, with penalties on the dodge roll if they're immobilized or downed.

The result is that critical hits tend to cap at around 25% of attacks against most enemies.

Gun dispersion now comes with a degree of random variance, both good and bad. High per and good vision/manip scores can reduce the penalty, while the bonus variance is constant.

#### Summary

Make it so that the player does not always crit.

#### Purpose of change

In DDA, if you had even a modest amount of skill, you would basically crit every melee or ranged attack against most enemies.

Furthermore, ranged damage and accuracy were almost entirely deterministic. Standing in the same spot shooting the same target in the same conditions, you'd basically always get the same hit rating and damage result. This made it impossible to crit in some situations and impossible not to crit in others!

#### Describe the solution

Melee: Melee crits used to only check the target's dodge score. This was a problem as many monsters, including bosses, had a dodge of zero. Now they check the higher of a monster's dodge or melee. The dodge check is bonused, but it can be reduced by effects such as immobilization, meaning dodgy monsters can use melee as a fallback skill, but it's usually worse.

Ranged: Firing a ranged weapon now adds a random bonus or penalty to the weapon's dispersion. The bonus is capped at a flat -200, while the penalty is determined by the shooter's perception stat times the average of their vision and manipulation limb scores. This penalty caps at 800. A character with 10 perception and 1.0 to their limb scores will reduce it by 400. The penalty can be reduced to a minimum of 100 if a character has 20 perception. Perception > 20 does not further reduce this penalty, but it can still buffer against limb score or stat penalties.

This bonus/penalty applies to monsters who use weapons too.

#### Testing

Monster Testing: Turrets are now sometimes not instantly lethal on the first turn, as it's possible for them to miss, especially when auto-firing. Feral security guards are about the same as ever, though they land grazing shots a bit more often.

Character Testing: A 10 perception character with 3 in all skills is still able to easily defeat small groups of zombies with a gun, though now sometimes they will waste shots or get lucky hits in. The same character with a Barrett M107A1 is able to bring down a juggernaut with a few shots. The same character with a compound hunting bow is able to one-shot zombies sometimes, though not all the time, and if the zombie closes to melee it is imperative that they get away and set up another shot, as grabs are disastrous for aiming.

Characters with 6+ skill and 16+ perception play more or less the same as before, but will still occasionally miss. Pobody's nerfect.

#### Additional context

It's possible the base damage on some ammunition, specifically bows, will need to be increased, but for now this is feels like a decent first step toward making the world more dangerous.

**NOTE: This makes glare more impactful! Wear shades or a hat!**

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -2554,7 +2554,7 @@ void Character::process_turn()
             for( const effect &eff : get_effects_with_flag( json_flag_GRAB_FILTER ) ) {
                 const efftype_id effid = eff.get_id();
                 add_msg_debug( debugmode::DF_CHARACTER, "Orphan grabbing effect found and removed from %s.",
-                       disp_name() );
+                               disp_name() );
                 remove_effect( effid );
             }
             grab_1.clear();
--- a/src/ranged.cpp
+++ b/src/ranged.cpp
@@ -2315,10 +2315,12 @@ dispersion_sources Character::get_weapon_dispersion( const item &obj ) const
     dispersion.add_range( get_modifier( character_modifier_ranged_dispersion_manip_mod ) );
 
     // Shooting is never completely deterministic.
-    // Superhuman (18 >) per can get better than +/- 200 variance.
-    double variance_min = 950.0 - ( std::min( 800.0, ( 40.0 * ( get_per() * ( ( get_limb_score( limb_score_manip ) + get_limb_score( limb_score_vision ) ) / 2.0 ) ) ) ) );
+    // Peak Human (18+) perception can get a bit better than - 200 variance, but it caps at -100.
+    double variance_min = 900.0 - ( std::min( 800.0,
+                                    ( 40.0 * ( get_per() * ( ( get_limb_score( limb_score_manip ) + get_limb_score(
+                                            limb_score_vision ) ) / 2.0 ) ) ) ) );
     dispersion.add_range( rng( variance_min, -200 ) );
-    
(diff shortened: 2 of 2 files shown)


anothersimulacrum:
> Furthermore, ranged damage and accuracy were [i]entirely[/i] deterministic. 

Damage yes (AFAICT), accuracy no. The dispersion sources are used as random sources, see `dispersion::roll`, used to generate the `missed_by` in `projectile_attack`. You've added a random random source to dispersion.

worm-girl:
> (This is for current DDA, I've been looking at [CleverRaven/Cataclysm-DDA#80151](https://github.com/CleverRaven/Cataclysm-DDA/issues/80151))

It's possible something in DDA's code has changed, but I did the ranged part of this PR to address my findings in https://github.com/CleverRaven/Cataclysm-DDA/issues/66609

anothersimulacrum:
Ah. "RNG is not strong enough to produce much variation in outcomes/the outcomes I want" is different from _entirely_ deterministic, but I see what you were saying. Thank you for the explanation!

worm-girl:
Amended the PR body to say "almost entirely" rather than "entirely" :P

Issue #395, opened 2025-02-23
Western boots encumbrance


neddy684:
Western boots have an unusually high encumbrance value (base 58 for well-fitting boots, 68 if too large). This does not seem intentional.

Also, there is a typo in the description.

![Image](https://github.com/user-attachments/assets/b23653cc-0727-428d-862c-eccf0635626a)

PR #398, opened 2025-02-24
Fix typo in boots_western and adjust encumberance

theFeu:
#### Summary
Fix typo in boots_western description and adjust encumberance

#### Purpose of change
Fix for boots_western, refs https://github.com/Cataclysm-TLG/Cataclysm-TLG/issues/395

#### Describe the solution
Fixed the typo and adjusted the encumberance to what DDA uses :)

### DIFF
--- a/data/json/items/armor/boots.json
+++ b/data/json/items/armor/boots.json
@@ -724,7 +724,7 @@
     "copy-from": "boots",
     "type": "ARMOR",
     "name": { "str": "pair of western boots", "str_pl": "pairs of western boots" },
-    "description": "Stiff leather cowboy boots with pointed toes and a low heel.  The wide cuffs could be used as as hoc pockets, or maybe even holsters.",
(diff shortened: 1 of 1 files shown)


worm-girl:
Per #166 and its related PRs, we're basically redoing all the assumptions about encumbrance, which means DDA isn't necessarily a good bellwether for our encumbrance values.

You are right that they were too high though. In this case it was because I'd erroneously left the "proportional" line in which was setting the encumbrance to an adjusted version of what was on the other boots and adding it to the base value for these boots before halving it when fitted. Which is confusing. So I just nuked the proportional line, set the encumbrance to what it was supposed to be, and left your typo fix in place. Thank you for jumping on both issues!

theFeu:
I see, thanks for the link!
I'll check out your docs about what your guidelines about clothing is for future PRs :)

Issue #388, opened 2025-02-21
Broken glass not consumed on (a)pply


mkikt4743:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

The game does not consume broken glass when placing it. BN fixed this and didn't cause the horrid lag that DDA has by keeping glass as an ammo item and creating a 'broken glass trap' recipe, where DDA fixed this by making broken glass into a non-stacking item again.

I am not personally aware of any way to do it myself, so I figured to make a bug report.

### Upload your save (optional)

[Kaupo-trimmed.tar.gz](https://github.com/user-attachments/files/18914618/Kaupo-trimmed.tar.gz)

### Steps To Reproduce

1. Load Save
2. (a)pply broken glass item
3. An item is not consumed from the broken glass stack, but the trap is still placed

### Expected behavior

Broken glass should be consumed when placed. It's really up to y'all whether to use DDA's workaround of making glass non-stackable again, or to use BN's workaround of making 'glass trap' recipe. If needed, I can link the recipe from BN. 

Or, if it's possible, the best solution is finding some way to consume items from the stack on (a)pply. As I've said before, I have not much idea on coding, but I love the project already and hope all the wonderful work you guys have managed continues! Thank you very much for everything y'all do.

### Game version

- OS: Windows
    - OS Version: 10.0.19045.5487 (22H2)
- Game Version: 1.0 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    The Last Generation [tlg],
    Alternative Map Key [alt_map_key],
    Slowdown Fungal Growth [no_fungal_growth],
    Stats Through Kills [stats_through_kills]
]

PR #410, opened 2025-02-28
Possible fix for broken glass issue

mkikt4743:
#### Summary
A possible fix for https://github.com/Cataclysm-TLG/Cataclysm-TLG/issues/388 , this uses Bright Night's method of making the broken glass into a trap item with a very short (1s) recipe, with the flavor slightly changed to say it is 'prepared to be scattered

#### Purpose of change
Bring forward a (potential) fix for the infinite broken glass. I personally prefer this because it still preserves the advantages of the glass being AMMO, such as lowering extreme haul and pickup times simulating the player character picking up every shard individually, as well as lowering lag when large piles of shards are propelled.

#### Describe the solution
A recipe, autolearned at level 0 devices, allows player characters to craft the AMMO broken glass into 'broken glass trap', which can then be (a)pplied to the same effect as broken glass had, while being actually consumed.

Could also revert broken glass back into being not AMMO.

#### Testing
I booted up a new world, grabbed some glass, tried to apply it, no luck. 
I crafted the new trap, the character knew the recipe with 0 devices, and it can be applied.
When zombies step on it, it makes some noise and deals minor damage. 

#### Additional context
image
image
With no trap in inventory ^, there is no option to apply the glass: V
image
image
image
With trap in inventory ^, the player character may (a)pply it to place the shards as usual: V
image

Let me know if anything further is needed. Happy to help, love the fork!


worm-girl:
This is a workable solution and I thank you for submitting it. Later today I'm going to check if I can just change the hardcoded trapfunc/iuse behavior to eat charges instead of count (should be easy). If I cannot, I'll go with your method and merge this PR.

worm-girl:
After messing with this for way too long, I have decided to just merge this PR so that the issue is at least temporarily patched. Sorry for the wait, thank you!

PR #432, opened 2025-03-04
Replace libz with built-in copy on macOS <10.15

a11ce:
#### Summary

The included libz is unreadable by macOS <10.15. In that case, we can replace it with a built-in copy.

#### Purpose of change

Let CTLG run on macOS 10.14 (and probably 10.13 because that's the compilation target, but I haven't checked).

#### Describe the solution

When launching CTLG for the first time, replace the included libz with a copy included in the OS. Then, write a sigil file so the replacement is only done once. 

It's theoretically possible that CTLG could directly use the OS copy of libz. I think this would make the process of building for macOS significantly different from other targets and generally seems like an avoidable can of worms.

#### Testing

Launched the game using the script from different states (including bad/missing libz) and made sure the sigil file prevents repeat copying.

### DIFF
--- a/build-data/osx/Cataclysm.sh
+++ b/build-data/osx/Cataclysm.sh
@@ -13,6 +13,12 @@ else
     K_FRAMEWORK_PATH=DYLD_FALLBACK_FRAMEWORK_PATH
 fi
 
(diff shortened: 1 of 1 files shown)


a11ce:
^ pushed to wrong branch, ignore

PR #435, opened 2025-03-04
Added chance to spawn both common and retail bionics to the High Tech…

sydknee-desilly:
… Lowlife scenario

#### Summary
Category "Brief description"
Added chance for both Common and Retail Bionics to spawn in the High Tech Lowlife scenario. The chance of Retail Bionics to spawn is lower to reflect their increased value and lower availability in an underground clinic.

#### Purpose of change
To add CBMs to the High Tech Lowlife scenario

#### Describe the solution
Changed the Basement_Bionic mapgen to include CBM Spawns

Making the spawn Static rather than chance based to reduce variance.

#### Testing
Started with the scenario High Tech Lowlife sever times to verify the spawns were working.


worm-girl:
It's not appropriate for both itemgroups to be used here (they're for totally different purposes) and it's also not appropriate to have one to nine bionics spawn in the room where you start.

worm-girl:
To be more specific: I think it would be best with two instances of retail bionic with no repeats and a 50% chance on both.

worm-girl:
Thanks!

PR #455, opened 2025-03-08
Fix enemies attack order

StrelokC:
#### Summary
Fixes an issue where enemies will always pick first attack in the list, which is always in alphabetical order causing enemies to always use same attack when possible f.e zombie always trying to grab player instead of trying to scratch them.

#### Purpose of change
Read below.

#### Describe the solution
Instead of iterating attacks list from the start, now it doesn't make it fully random, but just offsets attacks list by random number to start iterating from different attack (bite, grab, scratch) -> (scratch, bite, grab).

Make it fully random.

#### Testing
Skeletal juggernaut leaves our throat alone and picks other attacks.

### DIFF
--- a/src/monmove.cpp
+++ b/src/monmove.cpp
@@ -843,13 +843,35 @@ void monster::move()
 
     const bool pacified = has_effect( effect_pacified );
 
+    uint attacks_size = type->special_attacks.size();
+
+    std::vector<uint> random;
+    uint array_offset = rng(0, attacks_size - 1);
+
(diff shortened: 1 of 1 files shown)


worm-girl:
@StrelokC uint is not a standard type in C++ and cannot be used in Cataclysm. I'm not even sure how you got it to compile :P In the interest of making your life easier, I manually made some changes and re-opened the PR at #457 . It preserves your authorship in the commit log, thank you for tackling this.

PR #476, opened 2025-03-13
Fix incorrect dda->tlg in macro def

a11ce:
#### Summary
Fixes #468 

#### Purpose of change
The macro def flag `-DDATA_DIR_PREFIX` was replaced with `-tlgTA_DIR_PREFIX`. The DDA in this flag is not the name of the game.

#### Describe the solution
Change it back. I checked the rest of the repo for `-tlg`, this is the only sus occurrence.

#### Describe alternatives you've considered
Forking the compiler to recognize -tlg as a macro def that starts with DA.

#### Testing
@Reledia said it works.

PR #478, opened 2025-03-13
Renamed data/xdg files' names to match the actual name of the produced files

Reledia:
…d files

Renamed data/xdg files' names to match the actual name of the produced files

#### Summary
Category "Compiling"
The MAKEFILE was searching for the wrong files for the installation

#### Purpose of change
The MAKEFILE will be able to find the right files for the installation

#### Describe the solution
Change the name of the files inside the MAKEFILE

Noone

#### Testing
The compiling works with the command `make -j6 PREFIX=/usr PCH=0 CCACHE=1 RELEASE=1 USE_XDG_DIR=1 TESTS=0 LOCALIZE=1 LANGUAGES=all TILES=1 SOUND=1 install` and install correctly

#### Additional context
In order to work, it also needs the PR #476

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1147,9 +1147,9 @@ install: version $(TARGET)
 	cp -R --no-preserve=ownership data/help $(DATA_PREFIX)
 ifeq ($(TILES), 1)
 	cp -R --no-preserve=ownership gfx $(DATA_PREFIX)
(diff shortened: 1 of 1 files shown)


worm-girl:
This is the wrong solution. The correct naming scheme is cataclysm-tlg, so the other end of it needs to be fixed.

PR #490, opened 2025-03-14
Curb preposterous armor penetration on insect attacks

worm-girl:
#### Summary
Curb preposterous armor penetration on insect attacks

#### Purpose of change
Insects for some reason get to have a ton of armor penetration on their attacks, a power no other enemy type enjoys. This is unacceptable and it's weird that it's just this one enemy type. The giant wasp queen had 60 arpen on her sting and the armored centipede had ONE HUNDRED AND TEN on its bite. This is the epitome of lasagna design - layering a new mechanic over old unfixed systems that totally obviates them is not the way to go about things.

#### Describe the solution
Drop everything to like 10-20 tops. I don't really want one enemy type to have all this bespoke arpen when nobody else gets it, but it's not totally unreasonable for certain kinds of attacks - the player gets tons of arpen and nobody complains about that.

As a simple fix, for now I'll leave it in place and just bring it down to a reasonable level. Totally re-evaluating this stuff can come another time.

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -649,7 +649,7 @@
     "melee_skill": 6,
     "melee_dice": 3,
     "melee_dice_sides": 6,
-    "melee_damage": [ { "damage_type": "stab", "amount": 15, "armor_penetration": 25 } ],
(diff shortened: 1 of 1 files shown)


worm-girl:
Yes I know a centipede isn't an insect.

PR #493, opened 2025-03-14
Fix Night Vision

worm-girl:
#### Summary
NVGs give you a set amount of night vision, mutations add to what you naturally have.

#### Purpose of change
NVGs were just adding to your night vision, meaning people with night vision mutations etc were better at using them, which doesn't make any sense.


worm-girl:
I found a better way!

PR #576, opened 2025-03-27
Fix cmake "unused argument" errors in CI

johnaoss:
#### Summary
Noticed a bunch of builds failing with `clang++-17: error: argument unused during compilation: '-T LGTA_DIR_PREFIX' [-Werror,-Wunused-command-line-argument]`.
Looks like it got caught up in the DDA -> TLG renaming, but Cmake interprets the first two chars as the actual flag (this would result in `-T GTA_DIR_PREFIX`, which isn't referenced anywhere, but `DATA_DIR_PREFIX` is). 

#### Purpose of change
Hopefully fix the clang-tidy builds. Haven't tested but I'm hoping. 

### Describe the solution
Change the string back to `-DDATA_DIR_PREFIX` from `-TLGTA_DIR_PREFIX`

- rewriting the entire thing in a less janky build system
- installing several clang/cmake setups to build locally
- removing the workflows entirely

#### Testing
Relying on CI to test, fingers crossed it'll be greener.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,7 +182,7 @@ else ()
                 "${CMAKE_SOURCE_DIR}/src/prefix.h.in"
                 "${CMAKE_SOURCE_DIR}/src/prefix.h"
                 @ONLY)
-            add_definitions(-TLGTA_DIR_PREFIX)
+            add_definitions(-DDATA_DIR_PREFIX)
         endif ()
     else()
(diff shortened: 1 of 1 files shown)


johnaoss:
From `master`, sample workflow that hopefully would be fixed by this https://github.com/Cataclysm-TLG/Cataclysm-TLG/actions/runs/14115364057/job/39544127261

worm-girl:
Oh, thank you! Let's give this a go.

PR #579, opened 2025-03-28
Backport DDA 74561 - Add the Chiappa Rhino (but modify it)

worm-girl:
#### Summary
Backport DDA 74561 - Add the Chiappa Rhino

#### Purpose of change
Rhino good

#### Describe the solution
In addition to backporting the gun, I thought it might be appropriate to adjust some of its stats. The Rhino is notable for having its bore in a much lower position than a typical revolver. This helps with felt recoil, but because of the higher angle between the sights and the barrel, it's more difficult to aim. I have chosen to represent this by giving the gun 10 more sight dispersion, but a handling bonus of +1 (70 and 10 are the default assumptions for a handgun, so this works out to 80 and 11 in the gun's json file).

The result is a gun which is faster and easier to use, but less accurate than a typical revolver. Skilled shooters can overcome this, and gun mods can offset its penalties and boost its advantages. It's not the best gun in the world, but it's a nice little find, especially for weaker shooters with high skill and/or perception.

A regular backport. +2 handling instead of +1.


worm-girl:
Doh. Accidentally merged with +2 handling because I was messing around. I just committed a change to master to bring it back to +1.

PR #620, opened 2025-04-03
gun name standardize - pistol caliberer

dryramen2:
#### Summary
standardized gun names of pistol caliber guns, specifically 9mm, .40 s&w, .45, 10mm and .357 sig, maybe a draft version since I made this while sleep deprived
#### Additional context
peculiarities I've had working on this:
-calico m960 is described as both carbine and smg in description
-ptr 603 is a civilian variant of mp5 with only semi auto mode, but it's called submachine gun, so I renamed it smg. maybe should be named carbine?
-does the h&k briefcase count as an smg? I renamed it as such
-"semi auto mp 40 smg" name sounds odd, I think more than one company make reproduction of the gun? not sure if the name is appropriate as is
-sig pro sp2022, named shortened to just sig pro because the .40 version have shortened name also, and as of writing only the sp2022 version of sig pro is for sale by sig, so it's likely the only version of the gun around
-mac 10 is produced by military armament corporation and named m10 or m-10, but mac m10 smg sounded awkward, so I kept it as mac 10 smg
-"ba10 rifle" is the mp5/10????? who the fuck named this
I also only changed the "real" name of the guns and didn't touch the generic gun names, there might be issues arising from that
#### Testing
I typed in the name of some of the guns in debug spawner and it showed up, also spawned one


worm-girl:
This looks good, but I think we want to remove the calibers from the gun names, as they're displayed next to the names anyway. So the GunCo ManKiller 9mm and the GunCo ManKiller 10mm would both just be called GunCo ManKiller Pistol in the json, and the game would handle the rest.

worm-girl:
Thank you so much!

PR #636, opened 2025-04-08
A basic sample of how a documentation audit would look like.

Michael1993:


worm-girl:
I don't need a ruleset for documents that I am writing mostly by myself. The things that are needed are removal of irrelevant DDA documentation and updating stuff that has changed. I also do not think one sentence per line is the way to go here. I do not agree that it makes it easier to read, it just makes it read in an unnatural start-and-stop manner.

Michael1993:
Alright, that's mainly why I made this - wanted to know what'd be useful and what is not.

Michael1993:
I will make another attempt, no promises on timeline.

PR #645, opened 2025-04-13
Change tupperware containers to reduce spoilage inline with zipper bags.

Maepple:
#### Summary
Changes Tupperware containers to slow spoilage of food inside

#### Purpose of change
Make tupperware containers inline with zipperbags.

#### Describe the solution
Changed tupperware containers to have 80% of standard spoilage rate.

#### Testing
Game loads, tupperware containers in game show they spoil at 80% original rate


worm-girl:
Looks great, thank you!

PR #712, opened 2025-05-02
gun name standardized - numerical order

dryramen2:
#### Summary
standardized gun names in the following files
5x50
8x40mm
9x18
22
32
38
38super
40x46mm
44
44paper
45colt
46
50
57
84x246mm
270win
#### Additional context
the m3 carl gustaf serve a similar purpose to rocket launchers, so renamed as such


worm-girl:
Had to make some changes to the names as we don't need to say for example Rivtech RM2000 - the R in RM is already Rivtech. I handled these on my end though so we're good. Thanks!

PR #770, opened 2025-05-15
Update version.cpp

Alphavitto:
#### Summary
Instead of version, this file now shows build number of compile (I commented the version part, in the future the version could return and the build number could just be a separated entirely).

#### Purpose of change
Too many bug reports that include json error logs don't include the exact build number, which is the closest we have to the release versions. This is for testing purposes.

#### Describe the solution
Extract the build number from VERSION.txt file in the root folder of the game and have the error logs show it on the version line.

Scraping the release page of the GitHub.

#### Testing
Tested with a past version and it worked. This change only depends on whether the file VERSION.txt exists, if it doesn't exist, the build number should be "unknown".

#### Additional context

cataclysm-tlg-tiles_801rP6csFQ

### DIFF
--- a/src/version.cpp
+++ b/src/version.cpp
@@ -1,18 +1,36 @@
-#include "get_version.h" // IWYU pragma: associated
+// #include "get_version.h" // IWYU pragma: associated
+#include <fstream>
+#include <string>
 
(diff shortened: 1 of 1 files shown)


Alphavitto:
In the Windows version at least, when you compile the game, the VERSION.txt file gets generated.

PR #772, opened 2025-05-15
Update baseball player weapon proficiencies

calvinbeall:
#### Summary
Change the starting proficiencies of the baseball player profession to bludgeons instead of maces.

#### Purpose of change
Baseball players had the wrong weapon proficiency at character generation, since bats are bludgeons, not maces.

#### Testing
image
Made a new character with the baseball player profession. They had the correct proficiencies


worm-girl:
Thank you for doing this!

PR #780, opened 2025-05-16
Update crafting_gui.cpp

Alphavitto:
#### Summary
Added two modes (need to put the full name): [C]ategory and [S]ubcategory to the crafting menu filter. Also removed some unnecessary test.

#### Purpose of change
Because I felt like the crafting menu needed this for more complex filter searches.

#### Describe the solution
Just used source code definitions to come up with these filters. The source code has definitions for these categories/subcategories like CC_ARMOR or CSC_FOOD_SNACK but to the user that doesn't matter, so just decided to do it user friendly.

Only implement subcategory filter, but there are other times when you would want to filter whole categories too.

#### Testing
Works well! Even with edge cases.

#### Additional context
The text removed for difficulty range filter might be added again, since there's no other place where that was.
Supports bigger filters with other prefixes. (It's recommended to provide filter searches with category or subcategory first, since this will reduce the amount of time of the search to complete, since it doesn't have to see all of the items. This way using the d: prefix the search's faster)


Alphavitto:
Finally :D

worm-girl:
This is a really neat update, thank you!

PR #783, opened 2025-05-16
Remove dead mod code

worm-girl:
#### Summary
Remove dead mod code

#### Purpose of change
Removes some dead code related to mods we don't support, as well as a few references to portal storms.

#### Testing
Compiles and runs. Existing saves may experience error messages the first time they are loaded.


worm-girl:
They will not be re-added and I will not entertain any discussion about them.

PR #793, opened 2025-05-18
Add Charge chance on feral security guards flashlights.

Maepple:
#### Summary
Some feral security guards spawn with a heavy duty flashlight that doesn't roll for a battery charge chance unlike almost every other zombie.

#### Purpose of change
Make loot rolls more consistent between similar enemies.

#### Describe the solution
Added the charge chance roll to the loot drop in the json
#### Testing
Game runs, spawning feral secuirty guard flashlight and killing them now drop heavy duty flashlights with a battery that is not full.
#### Additional context
Inspired by the suggestion in discord.

### DIFF
--- a/data/json/monsterdrops/feral_humans.json
+++ b/data/json/monsterdrops/feral_humans.json
@@ -155,7 +155,7 @@
     "magazine": 100,
     "ammo": 20,
     "entries": [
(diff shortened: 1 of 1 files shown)


Maepple:
Noticed stun guns(tazers) in the same pool don't have charge associated.

PR #812, opened 2025-05-23
Fix grab not applying dodge penalties

worm-girl:
#### Summary
Fix grab not applying dodge penalties

#### Purpose of change
Grabbing was supposed to apply a -8 dodge penalty to the user (-5 for monsters, who have fewer sources of dodge bonus and cap out at 10) and prevent the victim from dodging at all, however this was found to be not working! The penalty seemed to apply to monsters, but not characters due to DODGE_CHANCE getting enchantified - the dodge_mod json object in effects.json was simply not affecting players.

#### Describe the solution
- Move dodge_mod for grab stuff to a DODGE_CHANCE enchantment. Hardcode (for now) the dodge penalty for grabbing or being grabbed to monsters, as they don't care about that enchantment.

DODGE_CHANCE could be made to work for monsters pretty easily, then that could be jsonized. Monsters with multigrab are currently sitting at -5 even with two grabs going. That seeeems fine?

#### Testing
- Did the shoebody bop to a couple of flesh raptors. Saw my dodge drop by 8.
- Got grabbed by a monster and saw my dodge go to 0.

#### Additional context
This might make grabs a lot deadlier for the player, we should keep an eye on this. Let me know if it's a problem, but for now we're just going to make it work like it was always intended.

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -3625,7 +3625,7 @@
     "show_in_info": true,
     "show_intensity": true,
     "//mods": "No swimmin' or manipulatin' with a held limb, other scores start severly debilitated and reach useless at about 50 grab strength",
-    "base_mods": { "dodge_mod": [ -10 ] },
+    "enchantments": [ { "values": [ { "value": "DODGE_CHANCE", "multiply": 0.0 } ] } ],
     "limb_score_mods": [
       { "limb_score": "manip", "modifier": 0.0 },
       { "limb_score": "grip", "modifier": 0.0 },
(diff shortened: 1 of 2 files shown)


seanicus:
Grabs are definitely feeling like a death sentence so far... on my 4th run this morning.  Perhaps the intended outcome?  I tend to play melee, so I probably need to adjust my strategies.

worm-girl:
Thanks for the feedback. Ideally you should be ok one on one-or-two but things should get worse as more enemies pile on. I can't check it out until tuesday or so, until then remember that armor and weapons with high to-hit and block power are your friend 😃

seanicus:
As I said, I probably need to adjust my style... I'm a "bo" guy all the way, so it should be the ideal weapon in these scenarios (+3, high blocking), but it's important to remember that the best defense against grabs will always be not getting grabbed in the first place.  Until your fork figures out a way to cancel my greed, stupid deaths will always be a thing.

PR #815, opened 2025-05-25
Update game.cpp

Alphavitto:
#### Summary
Now the activity distraction popup shows button "VIEW" when an enemy triggers the distraction.

#### Purpose of change
The "VIEW" button was visible when any distraction happened.

#### Describe the solution
Only show the said button when hostiles are dangerously close to you or simply spotted.

#### Testing
Might need some additional testing.


worm-girl:
Thanks!

PR #816, opened 2025-05-26
Gave stepladder ‘ALLOW_ON_OPEN_AIR’

Vegetabs:
#### Summary
Stepladder was causing an error when pushed/deployed from a higher z-level.

#### Purpose of change
Allow stepladders to fall a z-level without causing an error.

#### Describe the solution
Gave stepladder and aluminium stepladder the ‘ALLOW_ON_OPEN_AIR’ flag.


worm-girl:
I may want to change how ladders work, but for now this seems reasonable.

PR #819, opened 2025-05-27
Metabolism system rework.

Maepple:
#### Summary
Category "Brief description"

There is currently a issue with the metabolism system, especially when it came to mutations and bionics.
This aims to make the metabolism system more balanced and gives new functionality with a new enchantment effect.

#### Purpose of change

Currently there is issues with the metabolism system, which anything that lowers it effectively handicaps your weariness level making it difficult to do much with highly optimised metabolism reduction. It also currently is possible to become effectively immortal and not needing to eat anymore with cold blooded and recycler unit, as this brings the enchant value to -1 multiply (100% less), by design this also has far more value  as it also means your weariness no longer gets drained.

This PR reworks the system and changes light eater to conflict with cold blooded. (So -83% metabolism isn't possible through mutations alone)

#### Describe the solution

I created a new enchantment BIO_METABOLISM, and moved all current mutations into it, leaving recylcer unit CBM in METABOLISM. This allows modded content to still be compatible to affect general body rate but won't affect weariness if desired like body function metabolism is meant to do. BIO_METABOLISM is plugged into weariness threshold to be a better element to convey body function vs artificial means of impacting BMR. 

The recycler unit CBM is altered to give -0.25 MULTIPLY metabolism (Not bio), this gives 25% less metabolism of the current rate without effecting weariness. Chimera with 50% less calorie intake is broken and un-ideal. This change benefits high metabolism the most as it gives more benefit in extreme cases but far less in most cases compared to the old system.
All of these numbers is mutation + recycler synergies: Chimera gets 0.75 reduction in total rate where it was 0.5 before (New 2.25 vs old 2.5), very fast metabolism gives 0.5 (exact same as old system) and other higher metabolism has less benefit, an example is that rapid metabolism will net you 93% normal metabolism rate vs 0.83 of old system. 

This also allows making modded content like vampires easier, as it would allow you to have -100% metabolism, and still get weariness drain unlike the system this aims to change.

Metabolism will affect BMR without affecting weariness threshold, but bio_metabolism will affect weariness and BMR.

Json fix for this is possible but it would also essentially effect weariness threshold in places where I believe it shouldn't. This side would also have less impact for having it or not, having high metabolism + recycler would reduce the amount of calorie spent but not always in a matter that it seems like it should. It's a static number rather than being a general percentage.

Simply moving the enchant to BMR functions outside of where it is would make it so weariness threshold isn't affected to the degree intended by body rate metabolism. Changing weariness threshold alone, would still make it possible to get 100% less metabolic rate.

Changing how enchantment values are added up would most likely break alot of functions in the game unintentionally, especially in modded functions.

#### Testing

Game compiles, debug commands and making new debug mutations(not included in PR) to test with -1 multiply BIO_METABOLISM and METABOLISM, along with testing with Chimera mutation and lizard mutation, and some other related metabolism effects. The new metabolism system is working within expectations.

As of my testing it's currently not possible to get a metabolism rate of 0 without cheats or mods currently.

#### Additional context

Added a change of terms in the document on metabolism enchantment from BMI to BMR, to be more accurate. As well as documenting bio_metabolism.

If this system is not ideal and I forgot about something feel free to reject it.


worm-girl:
Closing this as it's not the sort of solution I want. I appreciate that you stepped up to take a crack at this, but I'd prefer to go the opposite direction and make everything simpler and with less overlap, rather than more complex.

PR #821, opened 2025-05-28
Fix reach attacks some more

worm-girl:
#### Summary
Fix reach attacks some more

#### Purpose of change
When I fixed different z-level reach attacks, I didn't realize that I'd inadvertantly broken kitty corner attacks on the same Z level.

#### Describe the solution
- Check the Z level of both the attacker and target and use std::round() when the levels are the same and std::ceil() when they are different. This ensures that Z-level attacks always round up while same-Z attacks do not.
- Ensure that this is implemented for NPCs and monsters as well, and that it works both with the (f)ire command and the tab key (autoattack).
- Disable proficiency learning during reach attacks.

- The option to disable circular distances needs to be removed as even prior to this change it wasn't really doing its job anymore, and there's no reason it should be there.
- I need to move the round/ceil thing to a single function so I don't have to repeat it in 3 places, but also *yawn*

#### Testing
Used a spear and a pike, let a grappler pull me off a roof, watched an NPC stab a guy kitty corner with a pike.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -10567,7 +10567,7 @@ std::vector<Creature *> Character::get_targetable_creatures( const int range, bo
                 }
             }
         }
-        bool in_range = ( is_adjacent( &critter, true ) || std::round( trig_dist_z_adjust( pos(), critter.pos() ) < range ) );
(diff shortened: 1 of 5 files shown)


seanicus:
"Disable proficiency learning during reach attacks." - curious about the reasoning behind this.  Not objecting, mind you, just curious.  Is it because grinding proficiency from behind the safety of a chain link fence is a bit too OP?

worm-girl:
People were going up on top of roofs and spam stabbing without any way for enemies to hit them back just to grind out the proficiency. It will get re-enabled when I am satisfied that this is no longer a safe activity.

PR #823, opened 2025-05-29
Update Alcohol Metabolism, Recycler Unit, Targeting System

worm-girl:
#### Summary
Update Alcohol Metabolism, Recycler Unit, Targeting System

#### Purpose of change
- The alcohol metabolism mutation wasn't really doing anything and didn't make much sense.
- The Recycler Unit CBM was hopelessly broken.
- The Targeting System CBM was conferring a gigantic bonus for free.

#### Describe the solution
- ALCMET now removes the lifestyle penalty for drinking alcohol (it still shows in the consumption menu, so as with Sweet Tooth.
- The Recycler Unit just reduces thirst. You could argue that this is basically the same as what the aero-evaporator does, but they're actually different in practice, and also shut up.
- The Targeting System and Recycler Unit CBMs are now toggleable and cost bionic power. This means NPCs can't use them at the moment, but they don't need the recycler unit anyway, so that's not too bad.
- Renamed Sugar-Loving to Snackaholic.

#### Testing
- Booted up the CBMs, saw they drained power as intended.
- Drank booze.
- Ate junk food.

#### Additional context
https://www.youtube.com/watch?v=hNUNx319UCM


Marrim:
Trogs having it is a Dwarf Fortress reference.

PR #827, opened 2025-05-31
Vision and Dodge Updates

worm-girl:
#### Summary
Category "Brief description"

#### Purpose of change
- Indefatigable and Night Vision were too good and were also redundant as per and cardio exist.
- It was trivial to get Dodge very high without much work.

#### Describe the solution
1) High night vision can read and craft in total darkness with some trouble. Full night vision does it no problem. 
2) NVGs can do both with some trouble. They see very brightly but they're horrible for detail work. Note that NVGs REPLACE your biological night vision, so you can't overcome this with traits.
3) The regular night vision trait is no longer available at chargen, except for mutants. It doesn't let you craft or read in total darkness, but will help in dim conditions.
4) The night blindness trait has been added to chargen. It reduces your NV range by 2, and can be removed by gaining a different vision mutation.
5) If it's too dark for you to see well, you take a -1 to your to-hit checks and a -20% to your dodge rolls. You will occasionally be warned about this if you miss an attack under these conditions, and it will show in your dodge on your @ screen. The check here is a bit more generous than the reading/crafting one. NVGs do fine with combat in the dark.
6) Indefatigable has been removed from chargen, except for mutants.
7) Indefatigable and Hyperactive have been stepped down from a 40% and 60% buff to a 20% and 40% buff. These mutations were not originally designed to stack with the cardio fitness system and were allowing players to get up to crazy levels of stamina with fast metabolism and an exercise regimen. They're still very, very good, keep in mind.
8) Dodge no longer gives 1/2 your dex as a bonus, it now gives 1/4 of your dex and 1/6 of your per. Yes, that's a debuff overall.
9) The dodge penalty for skating has been significantly reduced.

- Dodging needs to move away from a binary and become a thing that reduces damage on a partial success.

#### Testing
- Made a character, messed with crafting etc in the dark with various traits and equipment.


src/character.cpp
```diff
@@ -3011,6 +3016,30 @@ bool Character::practice( const skill_id &id, int amount, int cap, bool suppress
...
 }
 
+// The dumbest shit in the universe. This code exists to correct a deficiency in
+// the existing night vision code.
+// Normal Vision = 1.374320
+// Night Vision = 1.272479
+// High Night Vision = 1.155692
+// Full Night Vision = .934998
+// We want high/full NV to enable crafting in the dark, but they don't scale linearly, so here we are
+float Character::interpolate_night_vision( float vision_threshold ) const
```
RenechCDDA, line 3026: You should use [lerp](https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/e02b1f4b36956bfdfc09c419313526e11acfaf8d/src/cata_utility.h#L231) or [multi-lerp](https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/master/src/cata_utility.cpp#L267)

worm-girl: I have used Lerp a ton in the past, and somehow completely forgot it existed.

PR #859, opened 2025-06-12
Logarithmic melee stamina scaling

worm-girl:
#### Summary
Logarithmic melee stamina scaling

#### Purpose of change
Heavy weapons are badly disadvantaged in Cataclysm. Not only do light weapons do far too much damage (a separate issue), heavy ones offer no real benefit, are harder to use, take more inventory space/weight, and if DPS is the same, they suffer from a significant time wasting issue when they overkill a target compared to light weapons.

#### Describe the solution

Done:
- Move stamina from linear to logarithmic scaling. Normalize it at 20g instead of 16g, implying that 10 strength is average and not 8.

Considering:
- Give an even more favorable weighting to heavy weapons.
- Momentum tech for heavy weapons that gives an attack speed buff when you kill the target, similar to a D&D 3.x cleave.

- Sharply reduce the damage of basically all knives. The fact that a ka-bar does the same damage as a zweihander is absurd.
- Similarly reduce the damage of all lighter swords. The whole thing needs a rebalance a la #810 etc.
- Remove DPS readouts. Make sure new stamina readouts are accurate.

#### Testing
Ongoing

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -3250,19 +3250,21 @@ void Character::conduct_blood_analysis()
     }
 }
 
+
 int Character::get_standard_stamina_cost( const item *thrown_item ) const
 {
     // Previously calculated as 2_gram * std::max( 1, str_cur )
-    // using 16_gram normalizes it to 8 str. Same effort expenditure
-    // for each strike, regardless of weight. This is compensated
-    // for by the additional move cost as weapon weight increases
+    // using 20_gram normalizes it to 10 str. Heavier weapons
+    // get a discount on stamina burned/second.
     // If the item is thrown, override with the thrown item instead.
-
     item current_weapon = used_weapon() ? *used_weapon() : null_item_reference();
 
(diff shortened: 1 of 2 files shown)


worm-girl:
7daa445890e347b2f44ae25695fc73865a03ce7651a098f4914ed76c8b19ed19_1

LazyWizard:
`Momentum tech for heavy weapons that gives an attack speed buff when you kill the target, similar to a D&D 3.x cleave.`

Were you thinking along the lines of a hardcoded buff in the combat code, or an innate technique of the default brawling martial art? Barbaran would combine a little weirdly with the former as it gives you a block bonus on kills with a heavy weapon (so you'd be preparing for reprisal _and_ attacking relentlessly), while the latter feels kind of better than what the dedicated heavy weapon style gives you (but also IMO feels appropriate for brawlers).

seanicus:
I notice OSX assets aren't in the list of assets for this build... will they be forthcoming?  Not sure why I'm so eager, since I finally found a knife just in time for it to be nerfed...

PR #867, opened 2025-06-14
Fix OSX build

worm-girl:
#### Summary
Fix OSX build

#### Purpose of change
OSX build no work

#### Describe the solution
Remove some orphaned DPS code that was breaking the OSX build.


seanicus:
Thanks so much!

PR #884, opened 2025-06-23
Rebalance solar sensitivity damage

Maepple:
#### Summary
Category "Brief description"
Adds a check in the code that runs the damage checks for solar sensitivity and irritation to need <95% coverage

#### Purpose of change

Changes to many clothing and armour being reduced from 100% meant that solar sensitivity checks were trigger more often than they were originally intended.

#### Describe the solution

Adds a check in suffer.cpp to check for body part coverage and only trigger when it's <95% coverage

Scale damage according to coverage, as the game doesn't allow you to take decimal point damage, you can't scale 1 damage to affect the body part without changing how hit points work.

#### Testing

Downloaded a fresh copy of the branch, compiled, opened game, tested behaviour and working according to what I believe it should.

### DIFF
--- a/src/suffer.cpp
+++ b/src/suffer.cpp
@@ -1034,7 +1034,7 @@ void suffer::from_sunburn( Character &you, bool severe )
         float heavy_cumul_chance = heavy_eff_chance( exposure );
         float medium_cumul_chance = heavy_cumul_chance + medium_eff_chance( exposure );
         float light_cumul_chance = medium_cumul_chance + light_eff_chance( exposure );
-        float roll = rng_float( 0.0, 1.0 );
+        float roll = rng_float( 0.0, 0.98 );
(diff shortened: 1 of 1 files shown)


Maepple:
I just saw that I missed your reminder, on the discord channel, this is a quick suggestion of code, but I don't think it's 100% necessary, but changing out the if statement on the exposure, this would allow you to wear rioter masks to protect from the sun, I know balaclavas are 95% coverage, so without the change they would still work to protect effectively against the sun, lowering the head exposure doesn't necessarily change the face/mouth/eyes parts, so that's another consideration. 

Might save you on time if you decide to go with that. 

`if( exposure > 0.95 || ( bp == bodypart_id( "head" ) && exposure > 0.9 ) ) {`

PR #889, opened 2025-06-24
Fix armor blocking, damage, leather armor suit

worm-girl:
#### Summary
Fix armor blocking, damage, leather armor suit

#### Purpose of change
- Leather armor had the wrong encumbrance
- Boiled leather armor didn't make any sense
- A bunch of armor was missing BLOCK_WHILE_WORN and melee_damage values
- jeans_mod used knee pads in their recipe but don't have knee pads in their armor values

#### Describe the solution

Melee_damage: Roughly speaking, armor gets the following values according to what it's made of.
Leather/fur: 2
Chitin: 3
Wood/Plastic: 4
Acidchitin/bronze: 5
Steel: 6

BLOCK_WHILE_WORN was given to pretty much anything worn on the legs or arms which was designed to protect in combat or would be good for that kind of thing.

Increased the weight of most leather armor and updated its description. Adjusted the coverage and encumbrance on the suit of leather armor to be identical to its separates.

Flagged the suit of leather armor recipe as reversible, made it simpler and faster.

Added leather_boiled material, reworked the suit of boiled leather armor. It is no longer just "leather, but thicker", it's now its own material. It is a rigid material that has roughly double leather_treated's values, and is intended to be half as thick as any regular leather gear which is made of it. It's much lighter, slightly more protective/durable, and not as hot, but has a bit more encumbrance and zero breathability.

Currently there's only the full suit of leather armor, but a cuirass, arm guards, pauldrons, leg guards, helmet, and whatever else could be added at 2mm with recipes based on the existing one. The recipes for these items should not start with a finished leather cuirass (or whatever), they should be made out of leather sheets/hides/patches, boiled, and then put together. At that time, the existing full suit recipe should be tossed and a new one designed where you're just assembling separates.

#### Testing
Loaded in, all items seem OK.


Terrorforge:
Be advised that adding BLOCK_WHILE_WORN to leg armor does not redirect damage to the legs. You still get hit in the arms, so you don't get the benefit of the armor you're supposedly blocking with. You also need Unarmed skill as well as Melee skill to utilize the trait, which isn't a bug but is worth knowing for balancing purposes.


You can ignore this part because it's just my opinion man, but personally I feel like BLOCK_WHILE_WORN should be restricted, not expanded. It doesn't make a tremendous amount of sense to me, especially on leg armor and soft items, and it really devalues weapons capable of blocking and especially devalues martial arts with arm/leg block. It'd be one thing if it just redirected damage so you could wear heavy gauntlets and light armor otherwise, but blocking reduces incoming damage by ~%30-40 even at low skill levels, and the fact you just get that for free by wearing basically any armor in the game doesn't sit right with me. That should be a deliberate choice you have to make.

worm-girl:
I was under the mistaken impression that block_while_worn allowed you to try to preferentially use your armor when leg/arm blocking, but only if you already had leg/arm block.

Guess I just need to remove it as that isn't the case.

PR #922, opened 2025-06-30
Slightly nerf dodge bonus from dex

worm-girl:
#### Summary
Slightly nerf dodge bonus from dex

#### Purpose of change
Dex was previously giving 1 point of dodge score per 4 points of dex. This meant that an 8 dex character had +2 dodge and a 12 dex character had +3. Dodge was previously way, way overtuned and I've been trying to work it back to a reasonable level without ruining everything.

#### Describe the solution
You now get 1 point of dodge for every 5 points of dex. As before, you still get 1 point of dodge for every 6 points of perception. Overall this should drop most characters down 1 point of dodge, or none.

#### Testing
Compiles, runs, dodge scores look good.

#### Additional context
Dodge as an all or nothing thing needs to go away in favor of it becoming a scaling damage reduction like block, with only "full" dodges totally avoiding the attack. This is complicated by a bunch of stuff being exploded out to melee, monster, and mattack_actors. So for now a little nerf and we move on.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4031,11 +4031,11 @@ void Character::reset_stats()
 
     /** @EFFECT_STR_MAX above 15 decreases Dodge bonus by 1 (NEGATIVE) */
     if( str_max >= 16 ) {
-        mod_dodge_bonus( -1 );   // Penalty if we're huge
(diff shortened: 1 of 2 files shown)


seanicus:
@worm-girl I've noticed that the last two releases don't include OSX assets for download... any ideas?

PR #940, opened 2025-07-02
Remove fungaloids

worm-girl:
#### Summary
Remove fungaloids

#### Purpose of change
Fungaloids have been around since the Whales days, and they're just awful.
- Yet another humanoid alien relying on swarms of melee monsters
- Eat up a ton of CPU cycles wandering around doing nothing
- Shit up the map such that you just have to immediately drive away if you see them anywhere

#### Describe the solution
Mycus as a faction needs to be way more interesting, and that means it needs to function differently from all others on a fundamental mechanical level. I believe the pieces are in place for this, and we just need to do a little work. This is step one.
- Remove fungaloids and fungaloid shamblers
- Do not remove stationary fungaloids (towers, flowers, spires, etc)
- Do not remove fungalized monsters
- Remove "slowdown fungal growth" from the default mod list as it no longer does anything
- Replace an old fallback for missing monster IDs. Previously it was replacing missing monsters in existing save files with breathers. Now it puts a cat there. Nobody's gonna be unhappy about cats suddenly appearing in their save from time to time.

The way this faction is going to work is that the fungal furniture/terrain is going to be mostly harmless unless disturbed. At that time it will release spores. These spores get humans sick and eventually kill or harm them, and they fungalize zombies and some other creatures. These fungalized creatures then turn and fight other creatures. When they die, they spread spores. This way, we get much slower growth and situations that feel a lot more organic and fungus-y. The Mycus is a mycelial network colonizing via decay, piggybacking on zombies and arthropods to spread itself around. It is not an army of mushroom men.

#### Testing
Compiles, loads, runs. I visited a fungal area and it looked good.

image

#### Additional context
There's more work to do here. Fungal zombies ought to slowly create fairly substantial fungal patches rather than just poofing out a couple of spores when they die, but that'll take some time.


Agoelia:
Greatly appreciated change. If they were to stay as a faction I wouldn't complain. 

That said, one possible solution for your vision is having fungal creatures' corpses periodically emit spores until they rots (and give them spoilage time of x hours or days). It should also be possible to pulp or burn them to stop the spores. 

There should be at least a few hours between death and the first spores coming of the dead body. This would allow a survivor to not receive an instant death sentence if they encounter a fungal zombie and don't have a gas mask or antifungal drugs.

PR #962, opened 2025-07-05
fix protein bar size

dynst:
#### Summary
Bugfixes "Give protein bars a realistic size"

#### Purpose of change

469d695e5db3d543817f0ce8f61fcea917e7bd60 set the bar to 260 calories but didn't set the protein bar's weight and size to match.

#### Describe the solution

Give it a realistic caloric density with a weight of 50 grams, and a literal density that doesn't float on water.

Also while here adjust its vitamin contents to be proportionate with a ("No Nuts!" brand) nut-free dairy-free protein bar.

Set the pre-Cataclysm price of a single bar based on the same brand selling 12 bars for $35.

Making the protein bar not nut-free, or adding variants with peanuts or dairy in it.

Additionally adding spawns of `box_snack` full of these bars in stores or pantries or other locations it would make sense to find them.

### DIFF
--- a/data/json/items/comestibles/protein.json
+++ b/data/json/items/comestibles/protein.json
@@ -60,7 +60,7 @@
     "name": { "str": "protein bar" },
     "description": "A dense, chewy protein bar made from toasted grains with some fats and sweeteners to bind it all into a portable little snack.",
     "weight": "50 g",
-    "volume": "45 ml",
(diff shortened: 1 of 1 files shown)


worm-girl:
The vitamins were intentionally set to match a Clif bar. We can change the size/density, but the vitamins and kcal gotta stay the same.

PR #970, opened 2025-07-06
Update math_parser_impl.h to include missing dialogue methods

### LINKED ISSUE #967: Hallucinations causing crash
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Trying to craft after an early run-in with a Flaming Eye. The hallucinations appear to be causing a segfault and crash.

I found some DDA issues that seem relevant. [70654](https://github.com/CleverRaven/Cataclysm-DDA/issues/70654), which was alledgedly fixed by [70687](https://github.com/CleverRaven/Cataclysm-DDA/pull/70687). But 70687 merged in january of 2024, before the date of the TLG fork so idfk.

### Upload your save (optional)

[The Mindworld-trimmed-Hallucination-crash.tar.gz](https://github.com/user-attachments/files/21091582/The.Mindworld-trimmed-Hallucination-crash.tar.gz)

### Steps To Reproduce

1. Load the provided save
2. Attempt to craft a walking stick or something. Ignore the hallucinations.

### Expected behavior

no crash

### Game version

- OS: Windows
    - OS Version: 10.0.19045.5965 (22H2)
- Game Version: 2025-07-06-0918 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    The Last Generation [tlg],
    Slowdown Fungal Growth [no_fungal_growth],
    Bionic Slots [cbm_slots],
    Mind Over Matter [mindovermatter],
    Bionics Expanded [bionics_expanded]
]

worm-girl:
#### Summary
Update math_parser_impl.h to include missing dialogue methods

#### Purpose of change
fixes #967 

#### Describe the solution
Following what the crash log in 967 suggests, I updated thingie::eval() to have a better fallback which should avoid crashes even if it catches something it isn't designed for.

#### Testing
Loaded the save, crafted stuff, ignored the hallucinations, no problems.

#### Additional context
I think this might have been a mod-specific bug. I can only assume MoM was trying to run an EOC that was not prepared to deal with hallucinations. It's possible I was wrong about this fix as I didn't install MoM, but if that's the case then it's MoM's problem to fix and not mine.

### DIFF
--- a/src/math_parser_impl.h
+++ b/src/math_parser_impl.h
@@ -6,6 +6,8 @@
 #include <cmath>
 #include <string>
 #include <string_view>
+#include <type_traits>
+#include <utility>
 #include <variant>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Terrorforge:
... if it's an error with math functions and EOCs, it may not even be MoM, it might literally just be me. I made (what I thought was) a small modification to hastily backport a feature I like where it randomly pre-assigns you a number of paths, making it harder to learn new powers. It involved copying a pair of EOCs, which throw Math-related errors at first. I thought it was just syntax and I'd fixed it, but it's entirely plausible they were firing for hallucinatory NPCs and breaking everything. Oops. I'll have to do some testing tomorrow.

worm-girl:
If it's something on your end, this should still have fixed it, or at least it should downgrade it from a segfault to a simple error popup. Basically what was happening was some dialogue methods (you can see them listed in the code change) were not accounted for when the function was written, so when it unexpectedly got handed one it crashed.

Terrorforge:
Did the testing, and it doesn't seem to have made a difference. Inconsistent segfaults on both the old and new release. Weirdly, it seems a lot less consistent now, on both versions? Last night I could barely finish a few % of the craft before crashing, today I could sometimes craft three whole walking sticks without incident. Probably something on my end, who even knows.

Issue #62, opened 2024-11-21
Checklist 1: Bionic and Mutagen availability


worm-girl:
### Is your feature request related to a problem? Please describe.

This is a dev checklist to keep me organized. I want these all done before launch.

If you're testing/playing and this one isn't done, know that it's in the works. For now, a mutated or bionic start are probably the ways to go.

The following are very short term solutions. Longer-term or even medium-term I'd like to revisit these.

### Solution you would like.

- [x] Make hospitals reasonably likely to have autodocs
- [x] Make old labs more common and new labs less common
- [x] Make hospitals more common (they are weirdly rare)
- [x] Autodoc in the military bunker?
- [x] Put a few consumer-grade CBMs in electronic stores.
- [x] Up spawns of mutagens on dead scientists.
- [x] Wrecked transport van with an escaped mutant.

### Describe alternatives you have considered.

Longer term solutions:

- [ ] Small mutant faction that can hook you up.
- [ ] The final semi-aborted Old Guard quest should send you to deal with an Alpha mutant who orders you around but you can simply betray him and score some sick loot.
- [ ] More unique labs. See DDA #70771
- [ ] Make alpha mutagen more obviously tied to military stuff. Place some in military locations and make an alpha NPC or two.
- [ ] Update/improve mutant and bionic starts for better lore and to explain their place in the world.
- [ ] Unnatural Evolution trait: You pick a line at chargen and "evolve" into it over time. Greater version takes you over the threshold in about a year.
- [ ] Spontaneous Mutation trait: Lets you start with a mutation or two regardless of your profession.
- [ ] Jabberwock/Beast improvement expansion. Beast becomes a line that is about hunting and killing fusion-type enemies and devouring them to gain their power, all without setting foot in a lab. See #61 
- [ ] Eating significant amounts of mutant meat for prolonged periods should have a low chance to drop a mutation on you, unless you're adapted for it. Ditto for unclean water. These sources probably don't need to respect the stability rules.
- [ ] Very rare mad scientist lab room with a feral scientist who grabs you like the broken nursebot and sticks you with infusion before running off.

PR #983, opened 2025-07-09
Backport 82178 - Gradually mutate over time

worm-girl:
#### Summary
Backport 82178 - Gradually mutate over time

#### Purpose of change
#62 - It was a goal that we'd have starting traits available which gradually transform you into a post-threshold mutant. Someone at DDA went ahead and did it, albeit in a mod.

#### Describe the solution
- Change the flavor. It's not CRISPR and it's not a gland. You're just mutating for some reason. Maybe portal energies got you, maybe the blob just likes you, maybe you got experimented on.
- Remove the advanced mutant types from the list. Those are meant to be rarer and harder to acquire.
- Move it from a mod to mainline.

### DIFF
--- a/data/json/effects_on_condition/debug_eocs.json
+++ b/data/json/effects_on_condition/debug_eocs.json
@@ -1,0 +1,26 @@
+[
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_TEST_MUTATION_GRADUAL",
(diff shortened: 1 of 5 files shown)


RogueNova1:
we could make this portal storm related like how bombastic perks does it

PR #1001, opened 2025-07-11
Steel Mill Upgrade Project

RogueNova1:
#### Summary
Content "Steel Mill Overhaul"

#### Purpose of change

Make the Steel Mill a More Interesting Place

#### Describe the solution

Adding More Items to loot pool. as well as adding a Mini-boss Area

could possibly make it semi-dungeon crawl like. kind of how 7 days to die does there POI's

#### Testing

this will take multiple tests of different Jsons


worm-girl:
I wanted to do this, but I think we can absolutely have different kinds of steel mills. My plan was to make one where the blast furnace had become a sort of hell portal and the place was full of burnt zombies and special fire versions, so if yours is any different from that it'll be cool.

I do want to bring back magnetic zombies and make them part of the rust zombie line (zombie becomes magnetic, pulls a bunch of pieces of iron into itself, eventually evolves to be rusty) so you may want to add those to yours once they're ready. I'll bump up the priority on that so they're available to you.

RogueNova1:
image
here is some concept-esque stuff am working on my end. i believe there's a way to randomize terrain. at least i hope so because that would be nice to have.

anyway with the current floor tiles i have there is different thickness of slag. and it gets thicker the closer you get to the source

RogueNova1:
i think this is alright for a first semi-major addition on my end. this includes

1. New monster the "slag golem" a Slag encrusted zombie hulk. basically functions as a more armored version of a headless horror
2. rebalancing of the steel mill loot-tables which makes it more worth-while to loot them
3. a New Steel-mill Variant which has scorched zombies and a Massive Slag Spill

RogueNova1:
@worm-girl this PR is ready for your Review and potential merge

PR #1004, opened 2025-07-12
Improve chargen and cull old options

worm-girl:
#### Summary
Improve chargen and cull old options

#### Purpose of change
I noticed starting characters were a bit underpowered.

#### Describe the solution
Chargen changes:
- Single pool point buy is gone. It was never balanced, and while I initially left it in for simplicity's sake, I think the UI has been improved enough that it shouldn't be necessary.
- The stat pool has been increased by 1 for a total of 7! This allows you to make slightly stronger characters. 6 was a bit low.
- Starting skill points have increased from 2 to 4. This is accompanied by the removal of the free extra skill point when buying skills in chargen. That point never made sense for multipool and was fairly confusing besides.
- Some of the stat descriptions have been improved for clarity.

Defaults changed:
- Show Drug Brand Names - Now set to **true**. These were always a little inconsistent and there has long been some confusion over Xanax vs Valium in the generic system, and there's no good reason to use it imo. The best argument is that people from other countries might not know the drugs by American names, but the game is set in the USA, so we're going to call it what it's called. If Nethack players can learn the meaning of ELBERETH, you can learn what Vicodin means.
- Wander Spawns - Now set to **true**. This has not really had the Teleporting Monsters problem it's warned about for years. Any weirdness that occurs is almost always happening way outside of where you could see or even notice it, and static spawns have been default since TLG started. Note that this won't affect your existing saves as this option requires a world reset to take effect.

Options removed:
- Mutations by radiation - Lorewise, radiation excites the blob particles in your body and can trigger mutations. There is no reason this should be able to be disabled, though perhaps some in-game means ought to exist to prevent it (iodine or something? Probably also some traits.)
- Season length - This was not really supported by code at all and was basically a broken feature.
- Construction scaling - I don't see any reason for this to exist except to allow people to skip gameplay. Construction times exist to enforce kcal/weariness costs, advance zombie evolution, etc. and like crafting time is intended to be a tradeoff. For those who want to cheat or need to fix an error with the game, the debug menu exists.
- Auto notes map extras - Autonotes were largely removed. They will remain for stairs and favorited items, and may be introduced later for specific purposes, but the option is now deprecated.
- Event spawns (ie zombies wearing Halloween costumes during irl Halloween) - We have never had this feature here.
- Unload Reload and Shoot weapons - This was set to false by default, leading people to not know the feature even existed. It's also totally unnecessary, players can manually unload their weapon if they don't want to hold a draw.
- Item spawnrate - Lowering this was causing quests to bug out, and the game already has some pretty sparse spawns.
- Monster Speed and Monster Resilience - These wreak havoc on game balance in ways that aren't really linear like the options suggest, and this kind of thing would be better handled by lowering spawn rates, debugging better stats, etc.
- DISTANCE_INITIAL_VISIBILITY - ??? why was this ever an option??
Initial/max stat, trait, and skill points - Freeform is always enabled, so there's no reason for these knobs to exist. If the point buy settings don't suit you, go freeform!
- Skill/proficiency training speed - This is affected by intelligence. You can just raise your intelligence to increase this, or decrease it to lower it.

A few other obsolete options need to go, but not right now.

#### Testing
Loads, was able to make a character and play. No probems?


6DollarSuit:
its sad to see monster resilience and speed settings go, it was such an intuitive way to adjust difficulty, and the idea of replacing that with a need to debug the stats of each new character on the same world is a bit strange.

PR #1007, opened 2025-07-13
Make yoghurt spoil

Maepple:
#### Summary
Category "balance"

#### Purpose of change
Yoghurt doesn't spoil, which is unlike how it behaves irl

#### Describe the solution

Change the sealed spoiling rate of the container that contains yoghurt and pudding.

change the spoiling rate higher/lower

#### Testing

runs.


worm-girl:
So the problem with this implementation is that yogurt and pudding are different irl. Pudding will last years while yogurt will last like two weeks. I think we might need to give them different containers, because opened, they'll both go bad in a couple days without refrigeration.

worm-girl:
Adjusted the time a bit. Turns out sealed yogurt is good for longer than I thought. Also edited all instances of "yoghurt" in the game to use the more common American spelling "yogurt". Thanks for the contribution!!

PR #1009, opened 2025-07-13
Magnetic Zombies reborn

worm-girl:
#### Summary
Magnetic Zombies reborn

#### Purpose of change
Because #1001 is fast approaching, I thought it was time to bring back an old classic, the zombie magnetic pull ability. This PR adds two zombies:

Magnetic Zombie - A basic zombie that can pull magnetic weapons out of your hands.
Magnetokinetic Zombie - An upgrade that does an AOE magnetic pull which can be quite damaging to anything in the area, and also tries to pull magnetic weapons out of your hands.

Both zombies can also upgrade into rust or shocker types. They'll mostly only appear in special locations, but will be a valid upgrade option for the Master special ability.

To do:
- [ ] The AOE magnet pull could try staggering the player if they're wearing or carrying a lot of magnetic stuff.
- [ ] Increase range.
- [ ] Add option to make disarms from this and other effects create a popup message, so people don't unwittingly lose their weapon.

#### Describe the solution
image
image


RogueNova1:
well there is really to options on how we want to handle the Zombie spawn group for magnetics.
either option A. we merge them in with the Technician Spawns which would make sense

or Option B. if we want more fine grain control make a Separate Spawn group for the magnetics. kind of how Scorched zombies are handled

worm-girl:
Going with location-specific spawns as it's not being a technician that creates these zombies, but weird portal bullshit, same as shockers.

RogueNova1:
sounds good. once i have access to the spawn group i can rework some of the mapgens to include them as spawnables

worm-girl:
I would prefer to do that myself. I am not really looking for help with that sort of thing.

PR #1052, opened 2025-07-19
Patch flags and add fire_twohand to firearms

Maepple:
#### Summary
Category "Brief description"

This PR aims to fix some inconsistent behaviour, as well as add new behaviour to rifles/shotguns.

#### Purpose of change
Uh-
Worm wanted rifles to have fire_twohand flag, and ALL pistols and SMGs to have ALLOWS_BODY_BLOCK flags.
I added fire_twohand to shotguns as well since that made sense.

#### Describe the solution

Added the flags, reused inherited flags to make it more consistent and easier to do these changes in the future.
Added flags to the classes so they should be applied on a wider scope.

Hard coding the flags is a little bad as if you want to make a large sweep of changes into a item class it makes it take a long time longer to actually do that.
We also saw the problem that lead to this PR, being SW_610 acting like a handgun rather than a revolver because the flag "ALLOWS_BODY_BLOCK" was added without the rest of the revolver flags.

#### Testing

Game loads, Looking at the debug spawn menu, items seem to be much greater consistency in apply flags.

#### Additional context

This took me 7 hours

### DIFF
--- a/data/json/items/classes/gun.json
+++ b/data/json/items/classes/gun.json
@@ -70,6 +70,7 @@
     "copy-from": "gun_base",
     "type": "GUN",
     "name": { "str": "base pistol" },
+    "flags": [ "ALLOWS_BODY_BLOCK" ],
     "skill": "pistol",
     "sight_dispersion": 60,
     "//": "default handling for anything without a stock",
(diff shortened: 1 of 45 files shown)


worm-girl:
Woah, you went all out! Thanks!

PR #1061, opened 2025-07-21
Add json flags to lever action for reloading

Maepple:
#### Summary
Category "Lever Action Reload Flags"

#### Purpose of change

Make it more consistent between lever actions

#### Describe the solution

adds "RELOAD_ONE" flag to lever actions that were missing it.

#### Testing

Game runs, items now behave


worm-girl:
Thank you!

PR #1183, opened 2025-08-07
Fix minor error in cattail mission dialogue

aGreenJelly:
#### Summary
One of the missions where you gather cattails had dialogue that told you to gather 80 stalks; the actual amount required is 20.

#### Purpose of change
Make the dialogue match the mission requirements.

#### Describe the solution
Change '80' to '20' in the dialogue.

Not sure why there are two almost identical missions for cattail jelly. One of them could possibly be removed. For now I will just correct the typo.


worm-girl:
Thanks! I can't believe I forgot the dialog.

Issue #606, opened 2025-03-31
Bat Mutants love blood so much, it wraps around to soul-crushing despair


querns:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

When bat mutants drink blood, it's worth +15 morale. If they drink blood again in a short time frame, it's... another +15. You can keep doing this, until eventually the morale integer overflows to negative, and you become VERY sad for enjoying blood.

### Steps To Reproduce

1. Become a bat mutant with the relevant blood drinking mutations (Bloodfeeder).
2. Drink some delicious, delicious animal blood. (Try butchering an animal over a dumpster if you have problems wrangling the blood cleanly.)
3. Continue to drink blood until your morale overflows.

### Expected behavior

I'd expect the morale for drinking blood to have diminishing returns, like how most food works.

### Game version

- OS: Windows
    - OS Version: 10.0.22631.3880 (23H2)
- Game Version: 1.0 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    The Last Generation [tlg],
    Slowdown Fungal Growth [no_fungal_growth]
]

worm-girl:
#1228 has fixed this. Thanks @Maepple !

PR #1228, opened 2025-08-13
Give bloodfeeder tag max enjoyment

Maepple:
#### Summary

Bloodfeeder enjoyment of blood gave a static +15 morale, and was easy to get +60 and even overflow the value.

#### Purpose of change

Stops giving bats ability to get max enjoyment for eating or even overflowing into -60 morale.

#### Describe the solution

Add fun_max of +25 to hemovore_fun flags.

Changing the position of the consumption of the enjoyment malus of eating the same thing too much, but this would cause other issues and doesn't even fix the problem.

Max value of +25 seemed fair.

#### Testing

Game compiles on linux, game opens and creates a character successfully, giving the bloodfeeder mutation and consuming blood only allows you to get to +25 morale and won't exceed that.

#### Additional context

Saw the issue of this here:
https://github.com/Cataclysm-TLG/Cataclysm-TLG/issues/606

And thought I'd give it a try to fix.

### DIFF
--- a/src/consumption.cpp
+++ b/src/consumption.cpp
@@ -574,6 +574,7 @@ std::pair<int, int> Character::fun_for( const item &comest, bool ignore_already_
                 fun *= 1.1;
             }
         }
+        fun_max = 25;
     }
 
     // This cherry soda's just not the same...


worm-girl:
Oh, thanks! I hadn't thought of doing something this simple, I'm still trying to figure out how the overflow is even happening, but this is perfect in the meantime.

Maepple:
> Oh, thanks! I hadn't thought of doing something this simple, I'm still trying to figure out how the overflow is even happening, but this is perfect in the meantime.

I figure that setting a clamp limit for it somewhere would help on the whole but I don't know how the code interacts with other sections and I was mostly looking at how bloodfeeder was interacting. I'll see if I can find what could be causing the overflow though.

Issue #1091, opened 2025-07-25
Reinforced glass update unintentionally affects old labs


Xikorut:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Following the bash value changes to reinforced glass in #714, old labs have some issues regarding containment of creatures that should logically still be contained after the player shows up. The glass used for holding creatures in containment should not break when attacked by their intended occupants, such as mi-gos, zombie brutes, or shoggoths.

The values may need to be increased to cover these cases or the tiles swapped for a different type which has both the transparency and strength requirements.

In addition, the bash values for laminated glass are unchanged compared to CDDA, unlike the other forms of strengthened glass, which may be an oversight.

### Steps To Reproduce

Start in a lab scenario
Find one of the rooms with strong monsters contained by reinforced glass
Watch as they bash through said glass as if they just noticed they could after the game started

### Expected behavior

Containment cells should contain monsters they were designed to contain - such as zombie brutes, mi-gos, and shoggoths.

### Game version

- OS: Windows
    - OS Version: 10.0.19045.6093 (22H2)
- Game Version: 2025-07-24-0856 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    The Last Generation [tlg]
]

worm-girl:
They ought to be using ballistic glass for the mi-go and shoggoths. As for the brute, I think a zombie that bulked up in order to escape containment is a good setpiece...but if it's the same cell then it probably ought to have the same glass.

Xikorut:
I had similar thoughts regarding a further-evolved zombie being able to break the glass, though checking the bash values of the various monsters made it less realistic; none of the hulks or juggernauts seem to do more bash damage than a shoggoth with 7d5.

To make sense, it would require either removal of shoggoths from old labs or specific glass for shoggoth cells, though the latter would probably be less realistic than the builders simply using the same material for each cell instead of making shoggoth-only cells.

Ballistic glass has a str_min of 25, so it should at least hold mi-gos and brutes (4d6 and 3d8 respectively), so swapping the glass and removing shoggoths may be the most logical method that leaves the possibility of a zombie brute evolving into a zombie hulk (4d8) and breaking loose.

The finale room with six 1x1 cells may need auditing for its inhabitants as I am unsure what creatures it can hold (for some reason, it is harder to find than other finale rooms in the jsons).

worm-girl:
[...]

PR #1235, opened 2025-08-16
Upgrade old lab containment glass to ballistic

Xikorut:
#### Summary
Updates old labs to use ballistic glass instead of reinforced glass when containing monsters.

#### Purpose of change
Mi-gos and zombie brutes are natural spawns in various cells in old labs, but they are able to break through reinforced glass.
This is mostly intended to fix the overly common example of a room with a mi-go which immediately breaks the glass in its room in a direct line to a shoggoth spawn, leading to two loose shoggoths.

#### Describe the solution
Switches reinforced glass for ballistic glass for prisoner containment, nether creature containment, alien containment, and glass monster cage rooms. As a result, brutes and mi-gos are no longer able to break their glass while a brute that evolves into a hulk can.
Updates computer_session.cpp to support reinforced or ballistic glass for the netherworld access lab finale for backwards compatibility.

Updating all lab glass to be ballistic instead of reinforced, or more selectively doing so based on logic/hack difficulty.
For example, adding ballistic glass to military bunkers, bionics finales, weapons testing finales, portal finales, and possibly autodoc rooms (due to the difficulty of their hack).

#### Testing
Newly generated labs use ballistic glass instead of reinforced glass.
Able to open reinforced glass chambers from pre-existing labs.
Able to open ballistic glass chambers from new labs.
Mi-gos and brutes are unable to break through new glass.
Hulks are able to break the new glass (and old).

#### Additional context
When creating ballistic glass lab doors, I updated (and then copied) the regular ballistic glass doors to have values that match ballistic glass, similar to how reinforced glass doors match the TLG values for reinforced glass.

This fixes issue #1091 using worm girl's suggestion of using ballistic glass instead of reinforced glass.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-doors.json
+++ b/data/json/furniture_and_terrain/terrain-doors.json
@@ -72,17 +72,51 @@
     "flags": [ "TRANSPARENT", "DOOR", "NOITEM", "MINEABLE", "BLOCK_WIND", "SUPPORTS_ROOF" ],
     "open": "t_ballistic_door_glass_o",
     "bash": {
-      "str_min": 200,
-      "str_max": 400,
+      "str_min": 25,
+      "str_max": 90,
       "sound": "glass breaking!",
       "sound_fail": "whack!",
       "sound_vol": 20,
       "sound_fail_vol": 14,
       "ter_set": "t_mdoor_frame",
-      "items": [ { "item": "glass_shard", "count": [ 42, 84 ] }, { "item": "scrap", "count": [ 3, 5 ] } ]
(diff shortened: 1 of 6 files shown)


worm-girl:
I have already said no to this. It's based on a false premise. Mi-go and shoggoths cannot break windows of any kind as they do not bash. In the case of the zombie brute, he has evolved to be stronger since they put him in containment, so it is totally fine if he gets out. You are not supposed to be safe in labs and if a shoggoth eats you, that's what you get for being slower than a shoggoth.

image
image

PR #1305, opened 2025-08-30
Add documentation for Linux players (Debian, Fedora)

literategoblin:
#### Summary

Adds some documentation with installation instructions for the prerequisite dependencies and troubleshooting steps for Linux players.

#### Purpose of change

To reduce the amount of time needed for Linux users to start playing after downloading a release.

#### Describe the solution

Adds some documentation.

There is currently a pinned post and a FAQ channel in the Discord. These are fine but may not be the first place a player looks after discovering a problem. Also, not every player will join the Discord server.

#### Testing

N/A

#### Additional context

I only added Debian and Fedora instructions because those are the only distributions whose instructions are verified by another user (I use Debian, and teqa's instructions were for Fedora which is their distribution.)

Thanks to teqa and Avaa in the Discord, I based this work on theirs.

### DIFF
--- a/doc/LINUX.md
+++ b/doc/LINUX.md
@@ -15,5 +15,39 @@ sudo apt install libsdl2-2.0-0 libsdl2-ttf-2.0-0 libsdl2-mixer-2.0-0 libsdl2-ima
 ### Fedora
 
 ```bash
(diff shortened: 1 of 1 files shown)


worm-girl:
Thank you again for this!

PR #1310, opened 2025-09-03
HELICOPTER HELICOPTER

worm-girl:
#### Summary
HELICOPTER HELICOPTER

#### Purpose of change
Helicopters are kind of like helicrapters if you think about how useless they are. This is largely because their numbers seem pretty fudged and some major mechanics are not in place.

#### Describe the solution
- All single and two seater copters now have extra light frames.
- One of the medevac copters has an extra light frame. The other one is now called heavy-duty. It should be made a bit bigger, maybe to match the Blackhawk since they're often made of converted blackhawks.
- Dramatically reduced the fuel draw for the single and two seater copters. These guys are built for endurance and are supposed to be able to be airborne for 3 to 5 hours in some cases.
- Shrank the fuel tanks on said copters. They don't carry that much irl.
- Reduced the fuel consumption for blackhawk and medevac copters by about 33%. This brings them much closer to their IRL endurance times, which can be 2 to 3 hours in some cases.
- Reduced the power of the single/two seater engines, as these are not supposed to be very fast at all. Unfortunately there's only so low it can go before the things just won't take off.

- Ospreys and Apaches need the same treatment. Ospreys are probably about right but need to be carrying double the fuel they do now.
- The small rotor probably ought to have a bunch of air drag or something to slow the small choppers down.

#### Testing
https://youtu.be/xLgoGA9_JXE


5batek5:
maybe you will be interested in that ^_^:
https://github.com/CleverRaven/Cataclysm-DDA/pull/63761

PR #1335, opened 2025-09-06
Start removing black goo from zombie descriptions

worm-girl:
#### Summary
Start removing black goo from zombie descriptions

#### Purpose of change
- Every player alive either read DDA's design doc, despite it telling you not to do that, or they endlessly discussed the blob on forums or in chat rooms. This gave players the false notion that they know the lore, and that their characters know the lore.
- This was not helped by a bunch of zombies being described as being covered with or full of a strange black goo. Were that the case, any idiot could easily identify the problem.

#### Describe the solution
- Remove the goo.
- Remove zombie fursuiter and skull zombie. The former was a ridiculous joke and the latter was just not very interesting.

I doubt I got all of it, but this is a start.


worm-girl:
> Just FYI, the black goo is not blob or something, it's just a color of a rotten blood

It was initially written as blob, as evidenced by the old description for slime serum and several other sources. It was later retconned to just be rotten blood, which makes plenty of sense, but was misleading players who have some understanding that the blob is a sort of black goo.

PR #1343, opened 2025-09-08
Remove golden violin, change krar from gold to wood.json

Abexis:
Deletion of the golden violin item. Edited the krar from gold to wood.

#### Summary
Category "Brief description"
Deletion of the golden violin item, changes to the krar to account for it being wood instead of gold. 

#### Purpose of change
Cleaning up some legacy musical instruments and updating the krar (harp)

#### Describe the solution
Deletion of the golden violin item, changes to the krar to account for it being wood instead of gold. I also changed the weight of the krar to be in line with harps, roughly 15 lbs on the low end.

None.

#### Testing
Ran fine with no errors.

### DIFF
--- a/data/json/items/tool/musical_instruments.json
+++ b/data/json/items/tool/musical_instruments.json
@@ -258,51 +258,19 @@
     "melee_damage": { "bash": 12 }
   },
   {
-    "id": "violin_golden",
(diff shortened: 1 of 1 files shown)


worm-girl:
Thanks!

PR #1372, opened 2025-09-14
Innawood - fix foot armor recipes, NPCs spawning with modern items

Xikorut:
#### Summary
Fixes a missing extension in recipes/armor/feet.json so it actually uses the override recipes.
Fixes innawood NPCs so they no longer spawn with modern items.

#### Purpose of change
To enable the three leather/fur foot recipes to properly be accessible before smithing.
To fix modern items spawning on NPCs in innawood instead of innawood items.

#### Describe the solution
Rename recipes/armor/feet to recipes/armor/feet.json.
Add overrides to innawood for NC_NONE and NC_NONE_HARDENED which are used in npc::starting_inv_passtime for generating NPCs with gear scaled based on the game time.

Coming up with proper distributions for at least NC_NONE_HARDENED so later NPCs spawn with better innawood gear later on.
I am not an ideas/balance guy, so that exercise will be left to someone with better ideas than me.

#### Testing
Spawned many NPCs before and after.
Before, NPCs spawned with canned drinks and, at least once, an MP5.
After, NPCs spawned with appropriate equipment and no examples of modern tech.


worm-girl:
Thank you!

PR #1395, opened 2025-09-16
Gives separatory funnels more chances to spawn 

Sjb4243:
#### Summary
Separatory funnels are not very rare IRL, most labs will have at least one knocking around. 

#### Purpose of change
Separation is a bit of a hurdle for chem making so added more chances for them to spawn
#### Describe the solution
Added a low chance spawn in lab_glassware, as I can't see a reason that they wouldn't be there. 
Added a low chance spawn in chem_lab, as I don't know why they wouldn't be there either.
Added an additional chance in the science item group as well.
Adding separation to chemistry sets maybe? Or advanced chem sets which include a bit more goodies. idk this seems fine
#### Testing
No issues on making a new world with the updated jsons.
#### Additional context
None


worm-girl:
Looks good, thanks!

PR #1430, opened 2025-09-22
Add npc door behaviour overrides

andersonarp2810:
#### Summary
Add npc door behaviour overrides: don't walk through closed doors and leave doors open

#### Purpose of change
To ease micromanaging npcs behaviour. Currently, these two options are only toggleable through chatting directly with npcs. If you have multiple npcs following you, this change will make it easier to control them.

#### Describe the solution
Added override options on the Chat with NPC menu -> (o) Tell everyone on your team to temporarily...
Those two options are bound to the `w` and `a` keys.

The `c` and `d` keys were already occupied on this menu, so I figured it is better to not change them, to avoid player misinput from muscle memory.

#### Testing
Got a follower npc, set their miscellaneous behaviour to default, used Chat with npc menu to override them. Video below.

#### Additional context
I would like feedback on whether the `w` and `a` keys are good for this. Feel free to change them.
Also, I used the same string for these options as the one on dialogue, but to my surprise they are not automatically translated. Screenshot below the video.

https://github.com/user-attachments/assets/4d610f8b-f6ad-4aba-8ec4-88a8064f49c6

image

### DIFF
--- a/src/npctalk.cpp
+++ b/src/npctalk.cpp
@@ -589,6 +589,8 @@ enum npc_chat_menu {
     NPC_CHAT_ORDERS,
     NPC_CHAT_NO_GUNS,
     NPC_CHAT_PULP,
+    NPC_CHAT_AVOID_DOORS,
+    NPC_CHAT_CLOSE_DOORS,
     NPC_CHAT_FOLLOW_CLOSE,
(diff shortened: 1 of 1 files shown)


worm-girl:
Thanks!

PR #1501, opened 2025-10-02
Innawood - Remove duplicate veins file

N-Azurik:
#### Summary
Removes a duplicate veins file that was missed in #1497.

#### Purpose of change
It was causing any world with innawood enabled to not be able to load, because of duplicate definitions.
#### Describe the solution
Remove the duplicate veins file as it has now been replaced by the ore_veins file in furniture_and_terrain.
#### Testing
No error when loading into a innawood world.


worm-girl:
Thanks!

PR #1526, opened 2025-10-06
Lead volume and battery lead dissasemble rates

Maepple:
#### Summary
Category "Rebalance"

Reduces lead volume and increases it's weight

#### Purpose of change

A car battery when disassembled would produce 200L of lead, having about 9kg of weight.

Redoes lead to make in-game values follow real life.

#### Describe the solution

* Changes batteries to produce 60% of the battery by weight for lead,
* Changes the battery crafting values for lead.
* Changes the unit value of lead to 1ml, and increases it's weight to be 11grams per 1ml,
* Changes ingot deconstructs and crafts to use current volume values.
* Changes ammo lead craft and dissemble values to mostly adhere to weight to grain conversions, bullets more closely follow what they should be using to create.
* Reduce the amount of lead units to create solder wire.

Changing lead volume doesn't solve the issue of it being incorrect physically for lead mass, this change alters a lot of values to make it more closely mimic scientific data.

#### Testing

Game loads into world without issue, values are appropriately changed.

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -937,7 +937,7 @@
     "description": "A motorbike battery wired into a static power grid.",
     "item": "battery_motorbike",
     "breaks_into": [
-      { "item": "lead", "count": [ 69, 79 ] },
+      { "item": "lead", "count": [ 149, 159 ] },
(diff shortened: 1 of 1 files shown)


worm-girl:
Awesome, thanks!

PR #1534, opened 2025-10-08
Updates johnnycakes

Abexis:
Updates johnnycakes to be made of vegetables (corn) rather than wheat in the final craft.

#### Summary
Johnnycakes were made of wheat in the final craft rather than vegetables. This corrects that.

#### Purpose of change
Lets people with gluten allergies eat johnnycakes

#### Describe the solution
Changes the vitamins & material of johnnycakes from wheat to vegetable

None

#### Testing
Runs fine & without error. Wheat intolerant characters can now eat johnnycakes without negative effects.


worm-girl:
Thanks!

PR #1708, opened 2025-11-16
Backport DDA 83784 - Better eoc fix

worm-girl:
#### Summary
Backport DDA 83784 - Better eoc fix

#### Purpose of change
@akrieger seems to have found a better fix for the bug I tackled in #1238 , which should be less of a hit to performance than mine was.

#### Describe the solution
Revert my thing and backport his. Change static int reentrancy_depth to static std::size_t reentrancy_depth, something in DDA's codebase doesn't mind comparisons of integer expressions of different signedness, while ours does.


akrieger:
I have to fix the type still, but I dev on windows which does not warn on this.

PR #1739, opened 2025-11-22
Remove some boots, fix chitin

worm-girl:
#### Summary
Remove some boots, fix chitin

#### Purpose of change
- Chitin armor was supposed to be 3mm, a bit worse than plate mail but much lighter and easier to make. All pieces except the arm guards were 4mm.
- One or two pieces of the chitin sets had the incorrect melee damage.
- Chitin boots never made sense. There was no padding, no rubbery sole material, etc.
- Armored boots were redundant and a bit nonsensical.

#### Describe the solution
- Chitin boots -> chitin sollerets. These are easier to make and are boot covers, not entire boots.
- Chitin armor 4mm -> 3mm
- Remove armored boots
- Chitin armor is supposed to do 3 bash, biosil 4
- Chitinous -> chitin

### DIFF
--- a/data/json/items/armor/boots.json
+++ b/data/json/items/armor/boots.json
@@ -163,81 +163,108 @@
     "melee_damage": { "bash": 3 }
   },
   {
-    "id": "boots_chitin",
+    "id": "chitin_sollerets",
     "type": "ARMOR",
     "category": "armor",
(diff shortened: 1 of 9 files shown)


bean-b:
i also noticed that the chitin greaves seem to take longer to make then the whole chitin armor set does

Issue #1746, opened 2025-11-25
Auto Notes feature appears to be non-functional for map extras


Inglonias:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

DDA has an auto notes feature that is disabled by default that will automatically tag map extras for you. This feature appears to be non-functional for map extras (but it does function for dropped favorites and marking stairs)

<img width="599" height="80" alt="Image" src="https://github.com/user-attachments/assets/86d9b59c-6696-4ba5-96a5-deb7c4726acb" />
You can see in my settings that auto notes is switched on, but...

<img width="672" height="410" alt="Image" src="https://github.com/user-attachments/assets/a932082f-99d1-4318-94d6-1d72be2aeea1" />
When I go to the auto notes manager in the escape menu (esc -> 6) it is still off, and it appears to be stuck in the off position.

### Steps To Reproduce

1. Turn on auto notes in the settings.
2. Go into a game.
3. View the auto notes manager. Auto notes still appear to be disabled.

### Expected behavior

If you don't want to maintain this feature, that's fine with me too, but if it's there, I would expect it to work as it does in DDA.

### Game version

- OS: Windows
    - OS Version: 10.0.26100.7171 (24H2)
- Game Version: 2025-11-23-1507 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    The Last Generation [tlg],
    Inglonias Scenarios and Professions [Inglonias_Scenarios]
]

PR #1757, opened 2025-11-27
Remove auto notes manager from main escape menu

Inglonias:
#### Summary
Interface "Remove autonotes manager from main menu, as it is unsupported"

#### Purpose of change
In #1746 it was stated that this feature is no longer supported

#### Describe the solution
Remove the auto-notes manager from the main menu.

I considered tearing out _everything_ related to the auto-notes manager, but don't have the experience with this repository to do so at this time. Instead, I simply dummied the code out.

#### Testing
Built locally. Started a game. Viewed the main Esc menu. The option was no longer there to go to the auto notes manager.
image

#### Additional context
This is admittedly a half-measure. If the PR is rejected on those grounds, I understand completely.

### DIFF
--- a/src/action.cpp
+++ b/src/action.cpp
@@ -404,8 +404,6 @@ std::string action_ident( action_id act )
             return "open_options";
         case ACTION_AUTOPICKUP:
             return "open_autopickup";
-        case ACTION_AUTONOTES:
-            return "open_autonotes";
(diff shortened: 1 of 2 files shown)


worm-girl:
It's supported. It is used for the autonotes features which do exist.

PR #1839, opened 2025-12-23
Lower metal prices

doublescale:
#### Summary
Lower prices of some metals

#### Purpose of change
Lead to gold:
Smokes overvaluing a sack of lead

Since charges had been removed from some metals in commit 6104747df6414afae4b6d8218b1bacb55810d8f5 , their prices have been too high.

#### Describe the solution
Divide the prices by the old stack sizes, to a minimum of 1 cent.

Lead is an exception:
Its unit weight increased from 5g to 11g since that commit, so this change brings the pre-cataclysm price to 2 cents instead of 1 cent.

This change also lowers the post-cataclysm barter price of lead further, to 1 cent instead of 4 cents, to bring the material value of disassembled car batteries (a common way to acquire lead) closer to the assembled item's price.  The materials are still much more valuable, with the 1061 lead alone being worth 10.61$ vs. the 5.00$ car battery.  (The 13 sulfuric acid is worth 13$ extra.)

Set the prices to zero where they would've rounded to zero cents, and for lead.

#### Testing
Checked barter prices of lead and silver on the item examine screen, checked the price of a car battery's worth of lead in a trade with Smokes.

#### Additional context
Cents are not small enough as a monetary unit.


worm-girl:
Pretty much none of this shit should be worth anything to anyone. The plans for changing how trading works are so extensive that anything other than emergency fixes seems pointless at the moment.

PR #1840, opened 2025-12-23
Add missing category suffixes to gun names

doublescale:
#### Summary
Add category to gun names where they were missing.

#### Purpose of change
I found a DP-12 and didn't know what it was.

#### Describe the solution
I went through the guns where the category suffix ("... pistol", "... shotgun" etc.) was missing and added one.
I'm not sure I got them all right, I'm especially uncertain on "rifles" vs. "carbines", and on the ones in artisan_guns.json.

Adjusting the generic names too; in cases where there were variants, I only adjusted those.

Also, I'm pretty sure some of these guns don't follow the standard naming schema, but I'm not quite sure how that goes exactly.

#### Testing
Debug-spawned some of the affected guns, especially ones that had special singular/plural names.

#### Additional context
To find the guns with missing category suffixes, I flailed around with regular expressions for a while, before running this hacked-together thing in data/json/items/gun :
```bash
ag '"name": ' * | ag -v '(pistol|rifle|smg|mortar|carbine|revolver|gun|launcher|thrower|laser|cannon|flintlock)"'
```
There's probably a better way.

### DIFF
--- a/data/json/items/gun/artisan_guns.json
+++ b/data/json/items/gun/artisan_guns.json
@@ -4,7 +4,7 @@
     "copy-from": "mosin91_30",
     "type": "GUN",
     "name": { "str": "Obrez pistol" },
-    "description": "A cut-down, modified Mosin-Nagant.  It has a short barrel and the stock has been cut down and fitted with a mace head, with new low profile front and rear sights added.  The sides of the firearm have been reinforced with metal to withstand a beating.",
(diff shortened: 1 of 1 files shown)


worm-girl:
Looking good, thanks! I spotted issues with the Obrez and Pressin mostly unrelated to your changes and fixed them as part of this PR due to laziness.

PR #2047, opened 2026-02-02
Fix a seat in the budget town car with the wrong shape

gamma-delta:
#### Summary

The rightmost back seat in the budget town car is accidentally `left` shaped. This PR fixes it to be the right (heh) shape.

#### Purpose of change

It's a very minor cosmetic-only change, which is why I'm doing it as my first PR.

#### Describe the solution

Edit the typo in the json file

Live with the very tiny inconsistency

#### Testing

Not sure what to put here

#### Additional context

Hi Worm Girl


worm-girl:
Thanks!

PR #2069, opened 2026-02-06
make cargo shorts and pants pockets equal

andersonarp2810:
#### Summary
Cargo shorts had more pockets than cargo pants despite being smaller

#### Purpose of change
Give them equal pockets so i don't feel like i'm losing out by wearing cargo pants instead of shorts

#### Describe the solution
Copy pasted cargo pants "pocket_data" to cargo shorts
https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/b3e0bba873c6c0420301824538e999069f5b4e05/data/json/items/armor/legs_clothes.json#L875

None as i asked worm girl about which would be the best option beforehand

#### Testing
Opened item spawn menu, see cargo shorts now have the same pockets as cargo pants

#### Additional context
image
image

### DIFF
--- a/data/json/items/armor/legs_clothes.json
+++ b/data/json/items/armor/legs_clothes.json
@@ -1315,20 +1315,6 @@
         "max_contains_weight": "3 kg",
         "max_item_length": "165 mm",
         "moves": 100
-      },
-      {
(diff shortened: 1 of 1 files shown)


worm-girl:
Thanks!

PR #2087, opened 2026-02-08
Update zh_CN.po Chinese translations

CN-Xuan:
Update the Chinese localization catalog (lang/po/zh_CN.po). The PO file was modified (binary diff) to incorporate updated translated strings and localization changes.
[skip translation]


worm-girl:
Thank you so much!

PR #2111, opened 2026-02-10
Fix/chinesefix

CN-Xuan:
Corrected some incorrect translations in #2104

Added dozens of simplified Chinese translations.


worm-girl:
Thank you! These will be on Steam soon, I am just trying to fix bugs before I update there.

PR #2137, opened 2026-02-16
Allow lockpicking practice with vehicle cargo and door locks

gamma-delta:
#### Purpose of change

Currently you can only practice lockpicking with one type of lock

#### Describe the solution

Now you can take locks off of cars and practice with those.

From the description, it sounds like these are not normal car door locks, but more like a padlock you bolt onto the car. (It might be a non-tumbler lock, but it doesn't look like the game models that; otherwise lockers would need tubular lock picks.)

Cope


worm-girl:
I'm removing these weird car locks in favor of a better system, so while this is reasonable, there's no point merging it.

PR #2175, opened 2026-02-22
Port BN 7412 - Dynamic Atlas for dynamic tile recoloring

worm-girl:
#### Summary
Port BN 7412 - Dynamic Atlas for dynamic tile recoloring

#### Purpose of change
The game's tile drawing was fairly primitive. To get shadows, memory, night vision, etc. it was using bespoke preset texture layers. We can do better!

I was going to try to port BN 7412, which added DYNAMIC_ATLAS and a way to recolor tiles on the fly, but it turned out DDA and BN had moved too far away from each other to make that feasable, so while I'll still credit usagirei for the idea and initial work, I'm going to have to sort the implementation out for myself.

#### Describe the solution
Use the draw_options struct to optionally adjust RGB and HSV whenever we draw literally anything. This is a WIP but so far it is almost working!


worm-girl:
I'm gonna upgrade to SDL3 first. I'm not sure I'll even need this after.

PR #2182, opened 2026-02-23
Fix soaking long pole reverting to spear shaft instead of long pole

### LINKED ISSUE #2181: long_pole has no crafting recipe and only spawns at equestrian locations
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

The long_pole item is used as a component in several recipes (wooden pike, copper pike, steel pike, spear shaft, and a few others), but has only a single source in the game: deconstructing or bashing the f_showjump furniture found at equestrian/ranch locations.

It does not appear in any item spawn groups, is not a monster drop, has no mapgen loot entry, and has no crafting recipe.

The soaking_long_pole_active recipe exists and produces an item called a "soaking long pole" from a wood beam + oil — but it actually reverts to a spear_shaft, not a long_pole. The naming implies it should be a path to a long pole.

### Upload your save (optional)

_No response_

### Steps To Reproduce

Try to acquire long_pole

### Expected behavior

Should be able to craft or acquire this more easily

### Game version

latest

failboat78:
Closes #2181

## Summary

- `soaking_long_pole_active` was incorrectly set to `revert_to: spear_shaft`, meaning the crafting process (wood beam + oil) skipped `long_pole` entirely and produced a spear shaft directly
- This made `long_pole` effectively unobtainable through crafting — the only world source was deconstructing `f_showjump` furniture at equestrian/ranch locations
- Fix: change `revert_to` to `long_pole`, so the soaking process correctly produces what it says on the tin; players can then use existing recipes to craft it into a spear shaft or pike as desired

## Test plan

- [x] Craft `soaking_long_pole_active` (wood beam + oil, fab 3, SAW_W + CUT_FINE + FILE)
- [x] Wait 4 hours for it to cure
- [x] Confirmed it produces a `long_pole`, not a `spear_shaft`
- [x] Confirmed `long_pole` can be used in downstream recipes (spear shaft, wooden pike, etc.)

### DIFF
--- a/data/json/items/resources/wood.json
+++ b/data/json/items/resources/wood.json
@@ -90,7 +90,7 @@
     "price": "40 USD",
     "price_postapoc": "50 cent",
     "countdown_interval": "4 hours",
-    "revert_to": "spear_shaft",
+    "revert_to": "long_pole",
     "qualities": [ [ "HAMMER", 1 ] ],
     "melee_damage": { "bash": 18 }
   },


worm-girl:
Crap, I just fixed this without seeing this PR. Thank you for the submission, it's fixed now so this is unnecessary and I'm closing.

PR #2192, opened 2026-02-26
Add files via upload

TS-cara:
I’ve corrected some item names and fixed several overlapping creature names. Please feel free to delete the ja.mo file that was uploaded by mistake.


TS-cara:
The Japanese translation doesn't seem to be reflected in the Steam version yet. Did I miss a step or do something wrong on my end? Please let me know if I need to re-upload anything.

PR #2228, opened 2026-03-12
Fix NPC ebook reading: menu filter, activation, powergrid, cable, read-in-order

failboat78:
## Summary

NPCs assigned to read books had many bugs and "quality of life" issues that made ebook/tablet reading tedious.

### Fixes

**Menu filtering** — The "assign NPC to read" menu was showing unreadable books (too dark, wrong skill level, already mastered). The filter now uses `check_read_condition()` to show only books that are actually readable right now.

**Device activation** — When an NPC picks up an e-ink tablet or laptop to read from, the device is now turned on at activity start and turned off when the activity ends or is cancelled.

**Power consumption** — The reading activity now calls `ammo_consume()` at activity start to pay the device's power cost, matching how player reading works.

**Powergrid charging** — At activity start, if the device has a cable (`link_up` flagged), the actor now searches nearby tiles for a vehicle power port or powergrid  and plugs in, so the device charges while reading. Previously NPCs never plugged in at all.

**Cable disconnect / read-in-order loop** — The powergrid search used `map::points_in_radius` (Chebyshev/square distance) but `process_link` validates cable length with `rl_dist` (rounded Euclidean). Diagonal tiles that are within the Chebyshev square but beyond Euclidean range caused the cable to break loose immediately after plugging in. Fixed by adding an `rl_dist` guard to skip those tiles.

**Activity lifecycle** — Added a `canceled()` override so that turning off the device and unplugging the cable happen correctly on cancellation (e.g. interrupted by combat). Also fixed `revert_after_activity()` to call `activity.canceled()` for consistent cleanup.

**Activation order** — `do_npc_read()` now turns the device on *before* showing the book-selection menu, so `check_read_condition()` sees the device as active and doesn't filter it out.

**Read in order (ebooks)** — "Read books in order" was completing instantly when the NPC only had a tablet. Root cause: `item_contents::visit_contents` only recurses into `CONTAINER` pockets — `E_FILE_STORAGE` pockets (ereaders/laptops) are explicitly skipped. Both the reason-check in `can_do_activity_there` and the book-selection handler in the `NEEDS_BOOK_TO_LEARN` path now explicitly check `is_estorage()` + `ebooks()`.

**NPC-visible cable break message** — When a cable breaks loose the message was always sent as a player-only message. Now if the carrier is an NPC, nearby players see `Lara's e-ink tablet PC (on)'s over-extended cable breaks loose!`

These fixes apply to both e-ink tablets and laptops (any item with `E_FILE_STORAGE` and `link_up`).

## Test plan

- [x] NPC assigned to read a physical book → reads normally, no regression
- [x] NPC assigned to read a tablet (ebook) → device turns on, NPC reads, device turns off when done
- [x] NPC with discharged tablet near a vehicle or powergrid→ plugs in, charges while reading
- [x] NPC plugged in diagonally (1 tile diagonal) → cable stays connected, no immediate break
- [x] NPC plugged in, then moves out of cable range → player sees cable break message
  > `Lara's e-ink tablet PC (on)'s over-extended cable breaks loose!`
- [x] "Read books in order" with NPC that only has a tablet → picks up ebook correctly, does not complete instantly
- [x] Save and reload mid-read → device still on, activity continues correctly
- [x] NPC interrupted mid-read (combat) → device turns off, cable unplugs cleanly

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -1707,6 +1707,44 @@ void read_activity_actor::start( player_activity &act, Character &who )
     // starting the activity should cost a charge to boot up the ebook app
     if( using_ereader ) {
         ereader->ammo_consume( ereader->ammo_required(), who.pos_bub(), &who );
+        // Try to plug the ereader into a nearby appliance/powergrid to charge while reading.
+        // Search within cable length — any appliance tile in the network works.
+        if( ereader->can_link_up() && ereader->has_no_links() ) {
+            map &here = get_map();
+            const link_up_actor *link_actor = static_cast<const link_up_actor *>(
+                                                  ereader->get_use( "link_up" )->get_actor_ptr() );
+            const int cable_len = link_actor->cable_length == -1 ?
+                                  ereader->type->maximum_charges() : link_actor->cable_length;
+            bool plugged_in = false;
+            for( const tripoint_bub_ms &pt : here.points_in_radius( who.pos_bub(), cable_len ) ) {
+                // points_in_radius uses Chebyshev (square), but process_link uses rl_dist (Euclidean).
+                // Skip tiles that would immediately exceed max_length after plugging in.
+                if( rl_dist( who.pos_bub(), pt ) > cable_len ) {
+                    continue;
+                }
+                const optional_vpart_position ovp = here.veh_at( pt );
+                if( !ovp ) {
+                    continue;
+                }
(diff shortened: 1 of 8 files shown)


failboat78:
@worm-girl  tell me if this is touches too many files - I can break it up into smaller commits if that makes things easier.

PR #2232, opened 2026-03-13
Turn off active e-file devices when character falls asleep

failboat78:
## Summary
- E-readers, tablets, and laptops were left powered on when NPCs (or the player) fell asleep, draining battery overnight
- In `Character::fall_asleep()`, after cancelling any in-progress activity, walk all carried/worn items and call `transform_into` on any that have the `E_FILE_DEVICE` use action and are currently in their active/on state
- Covers `eink_tablet_pc_on` → `eink_tablet_pc`, `laptop_screen_lit` → `laptop`, and any future e-devices using the same `revert_to` pattern

## Test plan
- [X] Give NPC a charged tablet (`eink_tablet_pc_on`), let them go to sleep — confirm tablet is off when they wake
- [X] Same for laptop (`laptop_screen_lit`)
- [X] Player falls asleep with an active tablet — confirm it powers down
- [X] Battery charge on the device should not have drained during sleep

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -9526,6 +9526,17 @@ void Character::fall_asleep( const time_duration &duration )
             cancel_activity();
         }
     }
+    // Turn off any active e-file devices (tablets, laptops) when falling asleep
+    std::vector<item *> active_edevices;
+    visit_items( [&active_edevices]( item * it, item * ) -> VisitResponse {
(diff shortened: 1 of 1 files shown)


worm-girl:
Makes sense to me. You normally get prompted for stuff like flashlights, but there's no reason you'd really want to leave an e-reader on, so automatic shutoff is fine.

PR #2234, opened 2026-03-13
Add bathtub soaking mechanic for heat relief

failboat78:
## Summary

Bathtubs can now be used as a heat relief mechanic. Crouching or lying prone in a liquid-filled bathtub drenches the character, causing evaporative cooling. A morale boost is also granted when the character is actually overheating.

## What changed

**`data/json/furniture_and_terrain/furniture-plumbing.json`** — Added `SOAKING_TUB` flag to the bathtub furniture definition.

**`src/mapdata.h`** — Added `TFLAG_SOAKING_TUB` to the `ter_furn_flag` enum. Any future furniture that should behave like a soaking vessel can use this flag.

**`src/character.cpp` — `water_immersion()`** — Added a check after the existing water tile block. When a character is prone or crouching on a `SOAKING_TUB` tile, the liquid on that tile is summed. Volume thresholds reflect realistic bath depths (1 charge = 250 ml):

| Position | Required water | Effect |
|----------|---------------|--------|
| Prone | ≥ 30 L | Full body drench (excluding head), morale boost if overheating |
| Prone | ≥ 15 L, < 30 L | Partial lower-body drench, "too shallow" message |
| Crouching | ≥ 15 L | Lower body drench, smaller morale boost if overheating |
| Either | < 15 L | "Not enough water" message |

The bathtub spans two map tiles but liquid is stored per-tile via the keg examine action, so only the tile the character occupies is checked. The 15/30 L thresholds account for this — filling one tile with ~30+ L of water represents a realistic shallow-to-full bath.

Drenching provides cooling through the existing evaporative cooling system (`-0.008°C per body part per turn`, faster when already overheating). The morale boost (`morale_comfy`) only fires when the `hot` effect is active, so it rewards using the mechanic as an actual heat relief strategy.

## Test plan
- [X] Fill a bathtub tile with ≥ 30 L of water, go prone → body parts become wet, occasional "You soak in the water" message, morale boost visible if overheating
- [X] Fill with 15–29 L, go prone → "The tub is too shallow to soak in properly"
- [X] Fill with ≥ 15 L, crouch → lower body drenched, morale boost if hot
- [X] Empty bathtub or < 15 L → "The tub doesn't have enough water to soak in"
- [X] Standing upright in a filled bathtub → no effect

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -1333,10 +1333,6 @@ void avatar::set_movement_mode( const move_mode_id &new_mode )
         if( !instant ) {
             mod_moves( -100 );
         }
-        // Check for water immersion effects (e.g. soaking in a tub when going prone/crouching)
-        if( new_mode == move_mode_prone || new_mode == move_mode_crouch ) {
-            water_immersion( true );
-        }
     } else {
         add_msg( new_mode->change_message( false, get_steed_type() ) );
     }
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -13932,7 +13932,7 @@ bodypart_id Character::most_staunchable_bp( int &max_staunch )
     return bp_id;
 }
 
-void Character::water_immersion( bool immediate )
+void Character::water_immersion()
 {
     // Effects of being partially/fully underwater.
     map &here = get_map();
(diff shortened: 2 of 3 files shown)


worm-girl:
Looks good. I'll probably remove the messages later 'cause the player shouldn't need to be told they're wet if they made a decision to fill a tub and entered it. I also need to add this situation to is_in_water and stuff so it washes off acid and extinguishes fire, but I can handle that on my end in a day or two.

PR #2235, opened 2026-03-13
Fix stale item_location in efile activity causing save/load debug error

failboat78:
## Problem 1 — stale item_location on save

When a file is moved via \`EF_MOVE_FROM_THIS\`, it is removed from \`used_edevice\`'s pocket. The matching \`item_location\` in \`selected_efiles\` then holds a dangling safe_reference (the source item was destroyed). If the player saves at this point, \`calc_index()\` calls \`target()\` on the null reference, computes an out-of-range traversal index, and writes it to the save. On the next load, \`unpack(bad_idx)\` returns null and fires:

\`\`\`
DEBUG : item_location lost its target item during a save/load cycle
\`\`\`

### Fix

Two complementary changes in \`efile_activity_actor\`:

1. **Prevent new stale references** — in \`completed_processing_current_efile()\`, erase the matching entry from \`selected_efiles\` _before_ calling \`remove_efile()\`, while the item pointer is still valid.

2. **Tolerate existing stale references** — in \`serialize()\`, filter \`selected_efiles\` and \`currently_processed_efiles\` through a validity check before writing. Any already-stale entry is silently dropped rather than written as a bad index.

---

## Problem 2 — read menu shows files on unbrowsed estorage devices

Books/efiles stored on estorage devices (USB drives, tablets) appeared in the \`r\` (read) menu even when the device had never been browsed. The condition in \`read_inventory_preset::is_shown()\` fell through to \`!uses_energy()\`, which is true for USB drives, bypassing \`is_estorage_usable()\` — the only check that requires \`is_browsed()\`.

### Fix

Restructure \`is_shown()\` to always gate on \`is_browsed()\` when the parent is an estorage device, then check usability/energy as before.

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1497,10 +1497,19 @@ class read_inventory_preset: public pickup_inventory_preset
 
         bool is_shown( const item_location &loc ) const override {
             const item_location p_loc = loc.parent_item();
-            return ( loc->is_book() || loc->type->can_use( "learn_spell" ) ) &&
-                   ( !p_loc || ( !p_loc->is_estorage() || p_loc->is_estorage_usable( you ) ) ||
-                     !p_loc->uses_energy() ||
-                     p_loc->energy_remaining( p_loc.carrier(), false ) >= 1_kJ );
+            if( !( loc->is_book() || loc->type->can_use( "learn_spell" ) ) ) {
+                return false;
+            }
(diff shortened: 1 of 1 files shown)


worm-girl:
Only semi related and not part of this PR, but there's a new issue with efiles: 
Screenshot_20260313_190615_Discord

PR #2238, opened 2026-03-14
fix: prevent crash loading saves with NPCs reading books in nested containers

failboat78:
## Summary

- `item_in_container::unpack()` could dereference a null item pointer if its parent container failed to resolve (e.g. the owning character isn't in the critter tracker yet during load), crashing inside `item_contents::all_items_top`. Fixed by returning `nullptr` early when `!container`.
- `npc::load()` calls `clear_personality_traits()` → `unset_mutation()` → `reset_stats()` → `calc_focus_equilibrium()`, which tried to validate the NPC's `ACT_READ` book `item_location` before the NPC was registered in the critter tracker. Fixed by temporarily clearing the activity around the personality trait regeneration and restoring it afterward.
- `calc_focus_equilibrium()` accessed `activity.targets[0]` without checking if `targets` is non-empty, causing UB if an NPC had `ACT_READ` with no saved targets. Fixed with an `empty()` guard.

## Test plan

- [X] Load a save that has NPCs with `ACT_READ` activities (book in a bag in a backpack) — should load without crash or debug popups
- [X] Confirm affected NPCs continue their reading activity after load

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4166,7 +4166,7 @@ int Character::calc_focus_equilibrium( bool ignore_pain ) const
 {
     int focus_equilibrium = 100;
 
-    if( activity.id() == ACT_READ ) {
+    if( activity.id() == ACT_READ && !activity.targets.empty() ) {
         const item_location book = activity.targets[0];
         if( book && book->is_book() ) {
             const cata::value_ptr<islot_book> &bt = book->type->book;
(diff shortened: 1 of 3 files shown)


failboat78:
> Why are NPCs able to read books that they're not holding? Putting null guards in is a good thing, but that shouldn't be possible to begin with. Unless ACT_READ makes them take the book out.

Merge this one and I will do another PR to fix NPCs not actually holding the device.

PR #2261, opened 2026-03-16
fix: flour spawns as a proper bag instead of a single unit

failboat78:
## Summary

Two item groups placed `flour` using shorthand `[ "flour", prob ]` notation. Since `flour` has a default `container` of `bag_paper_powder`, this produced a bag containing only 1 unit (12.5g) — the same class of bug fixed in #2251 for `fertilizer_commercial`.

Fixed using quantity-appropriate bag groups to avoid flooding the player with flour:

| Group | Location | Old | New | Quantity |
|---|---|---|---|---|
| `sugar_house_kitchen_items` | Sugar shack kitchen | 1 unit | `flour_bag_paper_powder_1_70` | 1–70 units (partially used home bag) |
| `commercial_breakfast_mixes` | Supermarket shelf | 1 unit | `flour_bag_plastic_full` | 20 units (small sealed store bag) |

## Test plan
- [ ] Visit a sugar shack — kitchen should contain a partially-used bag of flour (not a single-unit bag)
- [ ] Visit a supermarket — breakfast mixes shelf should occasionally have a small sealed flour bag

### DIFF
--- a/data/json/itemgroups/Locations_MapExtras/locations_commercial.json
+++ b/data/json/itemgroups/Locations_MapExtras/locations_commercial.json
@@ -1312,7 +1312,7 @@
       [ "maple_sap", 50 ],
       [ "syrup", 100 ],
       [ "maple_candy", 10 ],
-      [ "flour", 30 ],
+      { "prob": 30, "group": "flour_bag_paper_powder_1_70" },
       [ "milk_powder", 30 ],
       [ "powder_eggs", 10 ],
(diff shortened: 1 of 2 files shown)


worm-girl:
Looks good. At some point I really need to go through and standardize grocery/supermarket shelves. The PR that backported the supermarket location attempted to create a new standard for this but it's really wonky, so now we're doing the XKCD comic about another competing standard.

PR #2407, opened 2026-04-21
Fix alcohol, mangle text when drunk

worm-girl:
#### Summary
Fix alcohol, mangle text when drunk

#### Purpose of change
image

#### Describe the solution
- Removed durations from drug effects which use vitamins.
- Vomiting no longer removes alcohol duration.
- Being drunk is now less debilitating.
- Removed "depressants" effect from most alcohol items. "depressants" and "stimulants" are being phased out. This means it's sort of hard to die of alcohol now, but that'll come back when we do the overdose pass.
- Text output is now mangled when drunk. I have added a number of safeguards to ensure that this only happens to ASCII text and that markup symbols and things around them are excluded. The degree to which everything gets messed up is based on how drunk you are.

#### Testing
- Text deformation (ascii-only) seems fine in every language I checked. Unfortunately it doesn't work in non-ascii languages, like Chinese and Russian. Sorry guys :(
- No save corruption or error messages after I implemented the ASCII-only check.


DSeyka:
Dunno about this drunk text feature, seems like it'd be a hindrance to user experience (as opposed to a character debuff) and could be easily misinterpreted as a bug.

worm-girl:
I on the other hand am quite happy with it.

PR #2473, opened 2026-05-02
fix practice recipe typo & bolt batch craft

LikeAShadow406:
#### Summary

stab practice requires stab skill not cut skill

remove batch craft from monster bolts

#### Purpose of change

make stabbing practice require stabbing instead of cutting

put monster bolts in line with other bolts

#### Describe the solution

cutting > stabbing

remove batch craft from monster bolts

#### Testing

made char with 2 cutting, did not have practice recipe, leveled stab to 2 via debug, had practice recipe

looked at bolt recipe in crafting menu, it looks proper

#### Additional context

this somehow took me 2 hours because I forgot something minor lol


worm-girl:
thanks!

PR #2492, opened 2026-05-05
Update: fix broken Russian translation

Ritchie997:
I hope I wasn't mistaken this time. I definitely sent the po file to git, not the mo.
Kick me if I made a mistake again.


worm-girl:
Seems good

PR #2507, opened 2026-05-08
Update: Russian translation

Ritchie997:
I waited a long time for my wife from the store, so I made a translation and updated it with more than 500 lines of text.


worm-girl:
Thanks!

PR #2514, opened 2026-05-09
Improve seed drill viability checking, streamline stooking

doublescale:
#### Summary
Fewer keypresses to set up grain stooks, accurate weather-checking on vehicle planters

#### Purpose of change
Setting up grain stooks for large harvests took many keypresses, even when using zones.

Planters on a vehicle were checking buckwheat's grow phases, regardless of the actual seeds used, and disabled the toggle in the vehicle-interaction menu without any explanation in case of inviability.
This allowed planting seeds late (like wheat) that would die from cold, and disallowed fast-growing plants (like oats) in some viable parts of the year.

#### Describe the solution
Remove the Y/N prompt when examining a stooking spot.  This had little utility since the action can easily be undone by deconstructing the active stook.

Vehicle planters can now always be turned on, and viability is checked when a specific seed is about to be planted.
If it is too early/late to plant, the seed is dropped to the ground (like in the case of trying to plant on non-plowed ground), and a message is printed like with manual planting if the player can see the planter vehicle part.

Maybe it's not appropriate to have seed drills magically check which seeds can survive being planted, like a survivor can; perhaps they should just be planted unconditionally, leaving the plants to die when planted out of season or underground (I did not test if that happens, though).

Also, it would be nice to have bulk actions to set up and take down stooks, perhaps with zones (constructing the stooks and moving the grain there can already be done that way).  A zone type to execute the examine-action on everything in it?

#### Testing
Happily set up a 16x12 rectangle of wheat stooks without prompts slowing me down.

Tried planting oats/buckwheat/wheat with my tractor during different times of the year (time skipped with debug commands), observed planting/dropping happening appropriately, did not see warning message when blinded.

And since the code comments mentioned caching the results of `warm_enough_to_plant`: Performance didn't seem awful (it shouldn't be any worse than doing all the planting manually).

#### Additional context
Diesel-powered farming is fun

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -3084,11 +3084,6 @@ void iexamine::stook_empty( Character &, const tripoint_bub_ms &examp )
     } else if( !here.is_outside( examp ) ) {
         add_msg( _( "The stook will need to be set up outside to dry." ) );
         return;
-    } else {
-        add_msg( _( "This pile contains grain that is ready to be left out for drying." ) );
-        if( !query_yn( _( "Stand the grain up to dry?" ) ) ) {
-            return;
-        }
     }
 
     here.furn_set( examp, next_stook_type );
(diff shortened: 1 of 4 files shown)


doublescale:
I'm not sure how to handle translations when removing strings;  should I remove those from the .po files too?

worm-girl:
> I'm not sure how to handle translations when removing strings; should I remove those from the .po files too?

No. Removing the string causes no problems, and I manually generate new .mos and .pos periodically.

PR #2515, opened 2026-05-09
Don't hide "spoils in" time on old food in eat-menu

doublescale:
#### Summary
Don't hide "spoils in" time on old food in eat-menu

#### Purpose of change
In the [E]at menu, with a sufficiently high cooking skill, there is a convenient "SPOILS IN" column that shows the time to spoil of food items, just like in the item's [e]xamine window.

However, on food that's getting old (considering *relative* spoilage, not an absolute time threshold), this entry was replaced by "soon!", forcing the player to check the examine-window for the actual time.
On old food, the row of the item is colored yellow, and the "FRESHNESS" column also shows "old", so I don't see the point in obscuring this information for extra warning-ness.

#### Describe the solution
Just print the time unconditionally.

Use the absolute spoilage time instead of the relative spoilage to consider food "old", since you'll still want to eat berries expiring in a day before eating pemmican expiring in 5 weeks, but fortunately this list is already sorted that way.  And I still wouldn't want the information to be obscured like this.

#### Testing
Before:
image

After:
image

Also tested this with a low-cooking-skill character.

#### Additional context
eating is important

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -796,7 +796,7 @@ class comestible_inventory_preset : public inventory_selector_preset
                 if( player_character.can_estimate_rot() ) {
                     if( loc->is_comestible() && loc->get_comestible()->spoils > 0_turns ) {
                         if( !loc->rotten() ) {
-                            return get_time_left_rounded( loc );
+                            return to_string_approx( get_time_left( loc ) );
                         }
                     }
                     return std::string( "---" );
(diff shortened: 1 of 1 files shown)


worm-girl:
It's like this intentionally. The problem here is that 5 weeks isn't "soon!". A fix would have to make "Soon!" appear if the spoil time is under 12 hours or something like that.

doublescale:
I could work on that.  Should the item-examine window show "Soon!" instead of a precise time in that case too?  My annoyance is that this table hides information from me that I can get with an extra [e] press and looking around in the less structured layout there.

worm-girl:
The intent behind "Soon!" is that there's no way a person could tell by looking at a sandwich that it was about to go off. I'd say that <12 hours should always say "Soon!" in every menu regardless of whether the item was flagged old or not.

PR #2523, opened 2026-05-10
Consistently hide food expiration time under 12 hours

doublescale:
#### Summary
Consistently hide food expiration time under 12 hours

#### Purpose of change
For characters skilled in cooking, the [E]at menu would replace the "SPOILS IN" value with "soon!" if the relative rot progress of the item was over 90%.
For items like pemmican, this could still be over a week until expiry.
Also, the [e]xamine menu would still let the player see a more precise estimate, getting around the information-hiding of that table.

#### Describe the solution
Stop showing the estimate in both views when the absolute time to expiry is under 12 hours.

Show "less than 12 hours" or "under half a day" for skilled cooks in the examination window in this case, to let players know the exact threshold without checking the code or reading the pull requests, but this can also be inferred by checking the last few estimates before the transition.

Alternatively, keep printing an estimate in all cases, but never drop to a finer granularity than 12 hours. This would be a slightly bigger change, since the fuzzy estimate printing is also used elsewhere, like for fire durations.  It might be better though, because it's questionable whether it's possible to distinguish 13 hours from expiry vs. 14 hours from expiry etc., as it is now.

#### Testing
Cooked liver (1 day shelf life), less than 12h but more than 10% from expiry:
image

Pemmican (3 seasons shelf life), 6 days (less than 10%) from expiry:
image

Pemmican, less than 12 hours from expiry:
image

#### Additional context
The fuzzy "less than 4 days" etc. formatting seemed busted during testing (flipping back and forth between "less than X" and "more than X" and "about X" with time going forward), maybe I'll investigate that code...

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -2030,14 +2030,10 @@ static std::string get_freshness_description( const item &food_item )
         }
     } else if( food_item.is_going_bad() ) {
         // Old food likewise is assumed to be fairly obvious.
-        if( player_character.can_estimate_rot() ) {
-            return string_format( _( "* This food looks <bad>old</bad>.  "
-                                     "It's just <info>%s</info> from becoming inedible." ),
-                                  to_string_approx( time_left ) );
-        } else {
-            return _( "* This food looks <bad>old</bad>.  "
-                      "It's on the brink of becoming inedible." );
-        }
+        // At this point, even skilled characters cannot tell the exact time of expiration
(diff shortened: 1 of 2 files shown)


worm-girl:
Looking good! Thanks for indulging me. It might be fun to make perishability estimates much looser and more tied to int and cooking skill, but that's a much bigger/less important project.

PR #2526, opened 2026-05-11
Fix sidebar weariness transition indicator range

doublescale:
#### Summary
Make use of the weariness transition indicator's full range.

#### Purpose of change
At weariness levels above Light, the transition indicator was using less and less of its full range.
Also, when weariness reached levels above Extreme, it kept jumping up.

#### Describe the solution
Instead of the widget using the "Fresh" weariness limit as the maximum and the weariness amount beyond the last threshold as the value, fix the maximum to 100 and calculate a normalized value in that 0 - 100 range consistent with the weariness level threshold logic.

There was a routine `Character::weariness_transition_progress` which seemed to provide something closer to the required data for this display, and maybe could've reduced some code duplication, but it was unused, and I think its transition range was off by one and didn't handle the looping levels going past Extreme, so I deleted that.

Also, picking a 0 - 100 range instead of 0 - INT_MAX or 0.0f - 1.0f is kind of arbitrary, but I don't think more precision than this is needed for any widgets.

Only by digging into this did I learn that the weariness level can go beyond Extreme, which caused some jumping in the transition bar which I always thought was kind of weird.  With this change, the transition gets clamped to 0 in that situation, so the player doesn't get information about how far beyond Extreme they are.  Still, I think this is better than the arbitrary-seeming flickering of the Transition indicator.

#### Testing
Took some meth and started working out to observe the Weariness going through its levels.
I saw previously unseen things:
image

Before, at the same weariness, one would have seen this, even though the character is very close to the "Moderate" level:
image

#### Additional context
The code around there is weird, but I didn't want to change much more than necessary.

Also, if I understand the level calculation loop right, the Fresh and Light weariness levels have the same range, and it only starts getting shorter at Moderate?  I'm not sure if that's intentional.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -5362,29 +5362,6 @@ int Character::weary_threshold() const
     return std::max( threshold, bmr / 10 );
 }
 
-std::pair<int, int> Character::weariness_transition_progress() const
-{
-    // Mostly a duplicate of the below function. No real way to clean this up
-    int amount = weariness();
-    int threshold = weary_threshold();
-    amount -= threshold;
-    // failsafe if threshold is zero; see #72242
-    if( threshold == 0 ) {
-        return { std::abs( amount ), threshold };
-    } else {
-        while( amount >= 0 ) {
-            amount -= threshold;
-            if( threshold > 20 ) {
-                threshold = static_cast<int>( std::round( threshold *= 0.75 ) );
-            }
-        }
-    }
-
-    // If we return the absolute value of the amount, it will work better
-    // Because as it decreases, we will approach a transition
-    return {std::abs( amount ), threshold};
-}
-
 int Character::weariness_level() const
 {
     int amount = weariness();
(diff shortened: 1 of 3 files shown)


worm-girl:
Most of the weariness system was spun up pretty quickly by one contributor who seemed extremely interested in taking a lot of code shortcuts, I think for performance reasons. It has made the system sort of inflexible and inscrutable, which led to nobody really following up on the initial work very much. This looks good to me, I can't think of any way it would break.

Also I am just now finding out weariness is uncapped? That seems, insane?

doublescale:
I'd question the performance of the code around there too, doing more iterations the higher the weariness value is, without bound (well, bounded by integer limits, I think they always terminate with that threshold==0 check at least).

I quickly looked into the first implementations of those routines, they seemed simpler then, so I assume they grew organically a bit with new requirements too.

If there really is no intended gameplay effect for the invisible weariness levels beyond Extreme (or for the weariness to even go that high), then it would make sense to terminate all these loops after 5 iterations.
It doesn't seem critical to do it right away though, I've never had problems with that;  replacing these routines in an overall simplification/overhaul eventually would seem more reasonable to me.

worm-girl:
My plan is to heavily restrict what the player can even do at the uppermost bounds of weariness, and I think the way to do that might be to add another tier where you're just fucking done, and put the cap in there.

PR #2538, opened 2026-05-14
Pre-final Update Russia

Ritchie997:
Hi, I have finally finished the translation to 100%, therefore I would like to ask to make an announcement with a request for help. I have translated an extremely large amount of text, but in places I noticed that sometimes incorrect, broken, or unsuitable translations slip through. I will still continue to work on the Russian translation, but I would like to ask Russian players: if they see a strange, unsuitable by description, or just unpleasant-sounding translation, to write to me about it either in our translators' chat or in private messages.


worm-girl:
I will make an announcement about this and post it in the changelog on Steam when it goes live there.

PR #2547, opened 2026-05-18
Fix

Ritchie997:
Correcting minor annoying incorrect words


worm-girl:
I have to close this without merging because you submitted multiples. I can only merge one update to the same file at a time.

PR #2569, opened 2026-05-24
Raise weight limit of XL plate carrier pockets

doublescale:
#### Summary
Raise the weight limit of XL plate carrier pockets holding ablative plates

#### Purpose of change
When trying to insert XL steel ballistic plates into an XL plate carrier, they would fall out immediately because they weigh 10.5kg, while the plate pockets were limited to 8kg.

#### Describe the solution
Raise the weight limit of the two pockets to 10.5kg.

Reduce the weight of the plates (but they probably had that weight for a reason).

#### Testing
Loaded up a big chimera character save with the changed JSON, was able to keep XL steel ballistic plates in his XL plate carrier, which would fall out before the change.

#### Additional context
It's weird that the insertion is allowed in the first place;  it took me a while to figure out that the plates had fallen out (in fact, dropped under my car before I drove off).

Also, the game UI does not show the volume/weight/length limits on these pockets, which made it extra confusing until I looked at the JSONs.


doublescale:
Actually, I noticed another problem;  there doesn't seem to be a way to craft the XL plate carrier currently.
In my run, I had crafted the old-style XL ballistic armor and got the new XL plate carrier via item migration (https://github.com/Cataclysm-TLG/Cataclysm-TLG/pull/2543/changes#diff-c85a522b809f1e00f504f4db1b6ae8ebb0de8d2ead382f695670c074a7c3f8d7).

But I'm not seeing a recipe for the new one getting added;  in fact, the recipes for the obsolete xl_ballistic_vest and xl_ballistic_vest_steel still seem to be in data/json/recipes/armor/torso.json .

worm-girl:
Thanks! I have fixed the recipe so it makes the proper vest now, and removed the old ones. The recipe is pretty clunky and regular sized plate carrier recipes don't exist, but you can just find those, so I'll do that later.

PR #2600, opened 2026-06-04
increase clay bullet result amount

andersonarp2810:
#### Summary
Increase clay bullet result amount. For comparison, one unit of rock creates 10 units of pebble. One lump of clay has the same volume as one rock. The recipe for lead sling bullets also results in 10 units.

#### Purpose of change
Increase the result of clay sling bullet recipe to 10.

#### Describe the solution
Add `"result_mult": 10,` to the clay sling bullet recipe json

None.

#### Testing
Before:
image

After:
image

#### Additional context
None.


worm-girl:
Thanks!

PR #2610, opened 2026-06-06
Fix consuming

worm-girl:
#### Summary
Fix consuming

#### Purpose of change
Consuming stuff was accessing item location in a weird way and also popping up a message inappropriately when you ate the last of your consumables. There was also a use-after-free with smoking.

#### Describe the solution
Guard against UAF in the smoking function, clean up the loc access and message popup in consumption so that it doesn't give a popup after you eat your last item, but does when you try to eat when you have nothing.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -14104,38 +14104,33 @@ bool item::process_fake_smoke( map &here, Character * /*carrier*/, const tripoin
 
 bool item::process_litcig( map &here, Character *carrier, const tripoint_bub_ms &pos )
 {
-    // cig dies out
+    bool is_joint = ( carrier != nullptr && typeId() == itype_joint_lit );
+    // It dies out.
     if( item_counter == 0 ) {
         if( carrier != nullptr ) {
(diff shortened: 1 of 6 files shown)


src/item.cpp
```diff
@@ -14104,38 +14104,33 @@ bool item::process_fake_smoke( map &here, Character * /*carrier*/, const tripoin
...
-                on_drop( carrier->pos_bub(), here );
-                carrier->i_rem( this );
-            }
+        const bool had_transform = ( type->transform_into.operator bool() );
+        if( had_transform ) {
+            auto transform_item = type->transform_into.value();
+            transform_item.transform( carrier, *this, true );
         } else {
             type->invoke( carrier, *this, pos, "transform" );
-            if( !this->is_null() ) {
```
vaz, line 14121: did we lose this guard when this was rewritten?

re: #2690

worm-girl: This guard was re-added but it seems there's another issue. Hm.

PR #2636, opened 2026-06-11
Better starting location search

hom-jesus-sa:
#### Summary
Search for any valid terrain, not just one when picking starting location. Also limits z-level search to z-levels within location constraints. Search occurs in a multi-ring method now too, which can be expanded later

#### Purpose of change
Current searches for start location can fail and afterwards they try again by genning new overmaps. This is fine when looking for one specific terrain, but it is inefficient for the purposes of looking for multiple valid terrain types since new overmaps must be generated for each terrain that is searched over.

#### Describe the solution
I added a minor thing to speed up search (only look z-levels where a desired location is valid) with start_location_placement_constraints.allowed_z_levels. Added another detail to allow further granularity of search (can specify a number of rings to search through before giving up on a search).

"For lake-dependent scenarios, we don't need to fully generate an overmap to know if it has a lake in it because the noise for generating lakes is independent. For a large body of water spawn, we could quickly narrow down overmaps to generate by only looking at places where the lake-noise would allow a lake. We could add an optional "feature_hint" param or something to the start_location JSON allowing us to quickly pare down which overmaps need to be generated for particular starts."
^^^ I didn't implement this yet but could do so in the future.

#### Testing
I built the game and tested a "Call of the Deep" scenario per forum poster Flesh Forge:
  {
    "type": "scenario",
    "id": "call_of_the_deep",
    "name": "Call of the Deep",
    "points": 0,
    "description": "At the start of the Cataclysm, you find yourself beside a large body of deep water.  Are you drawn to it?  Are you fleeing from it?",
    "start_name": "Beside the Water",
    "allowed_locs": [ "sloc_lake_shore", "sloc_cabin_lake", "sloc_cabin_lakeside", "sloc_lighthouse_ground", "sloc_lighthouse_small_ground" ],
    "flags": [ "LONE_START" ]
  }
It worked immediately and my generated character spawned by a lake. I didn't test all scenarios.

### DIFF
--- a/src/start_location.cpp
+++ b/src/start_location.cpp
@@ -86,11 +86,6 @@ int start_location::targets_count() const
     return _locations.size();
 }
 
-omt_types_parameters start_location::random_target() const
-{
-    return random_entry( _locations );
-}
-
 bool start_location::requires_city() const
 {
     return constraints_.city_size.min > 0 ||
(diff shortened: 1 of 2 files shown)


worm-girl:
Thanks so much! Let's merge and give it a whirl. If it works I'll have to check out how quest locations are being searched.

PR #2640, opened 2026-06-13
Build "Fix macOS build warnings and clean up Makefile"

zacbrown:
#### Summary
Fix macOS build warnings on Tahoe (26.x), dangling Makefile targets (`make test` etc), and update .gitignore to ignore intermediate build artifacts and the final binaries.

#### Purpose of change
macOS builds emit `-Werror` failures and `make clean` is broken after the test suite was removed.

#### Describe the solution
- `math_parser*.cpp`: route always-throwing lambdas through a new `[[noreturn]]` `math::raise` helper to satisfy `-Wmissing-noreturn` (lambdas can't take the attribute directly).
- `npctalk.cpp`: fix `string_format` operator issue.
- `Makefile`: drop the dangling `tests`/`check`/`clean-tests` targets and `-C tests` recursion left over from test-suite removal, fixing `make clean`.
- `.gitignore`: ignore macOS build outputs (`cataclysm-tlg[-tiles]`, `zzip`, `*.dSYM/`).

Marking the lambdas `[[noreturn]]` directly — rejected, invalid in C++17 (attribute applies to the type, not the function).

#### Testing
`make TILES=1 SOUND=1` compiles clean (no `-Werror` warnings); `make clean` succeeds.

#### Additional context
None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,8 @@ cata_test
 cata_test-tiles
 cataclysm
 *cataclysm-tiles
+*cataclysm-tlg
+*cataclysm-tlg-tiles
 cataclysm-vcpkg
 cataclysmdda-*
 chkjson*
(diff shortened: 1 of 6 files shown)


worm-girl:
Thanks! Let's give it a whirl.

PR #2650, opened 2026-06-14
fix not being able to interact with vehicles unless you are playing an instrument

deltanedas:
#### Summary
title

#### Purpose of change
i dont think you should need to be playing a guitar to be able to steal parts from cars etc

#### Describe the solution
removed an incorrect !

maybe im too grumpy and mechanics should always be playing music..?

#### Testing
in my mind

though something else is broken with it as on master:
- spawning a harmonica with holder
- wielding and playing it, seeing i now have the status effect
- trying to remove a car battery
still doesnt let me steal a car's battery, so i imagine it would just be dead code if this is merged until that is fixed

#### Additional context

mistake from https://github.com/Cataclysm-TLG/Cataclysm-TLG/commit/7cff8f7bf863298a7d99b83868f1a29de4b69dd1

### DIFF
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -651,7 +651,7 @@ task_reason veh_interact::cant_do( const map &here,  char mode )
     bool enough_light = true;
     const vehicle_part_range vpr = veh->get_all_parts();
     avatar &player_character = get_avatar();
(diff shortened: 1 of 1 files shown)


worm-girl:
ah shit i pushed a fix before noticing this. Thank you for submitting though!

PR #2673, opened 2026-06-19
Fix treetop overlays

worm-girl:
#### Summary
Fix treetop overlays


worm-girl:
Gonna do this a better way

PR #2712, opened 2026-07-01
Adding strength enchantment for permanent strength effects like mutations

### LINKED ISSUE #190: Strength Mutation HP/BMI/Carry Weight
### Describe the bug

#189 broke strength mutations by enchantifying them, which removed their HP bonus, BMI adjustment, and probably other stuff. I need to make a second kind of strength enchantment that _does_ raise those things and make sure all the appropriate mutations use it.

### Attach save file

_No response_

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Versions and configuration

_No response_

### Additional context

_No response_

aattss:
#### Summary
Fixing mutations that improve strength to improve things that base strength affects like max hp and BMI.

#### Purpose of change
An attempt to fix #190. When changing mutations that increased strength to use enchantments, it stopped affecting a couple values like health and max BMI, because the enchantment system is used to support effects like exo suits or magic potions or such where it isn't actually changing your body. But the mutations that affect strength are usually supposed to be changing the character's actual physical body, so the prior behavior was preferable. Or at least that was the explanation recall seeing in the stale issue CleverRaven/Cataclysm-DDA#72698 when I was looking into the CDDA issues.

#### Describe the solution

As suggested in #190, I created a new enchantment, which I decided to name STRENGTH_NATURAL, and plugged it into the three places I saw get_str_base being called where the other 3 stats weren't also being used, and then added it to the mutations with the STRENGTH enchantment where I thought it made sense. To clarify, this is in addition to STRENGTH. If STRENGTH_NATURAL is used without STRENGTH then it won't increase the strength calculations that are impacted by STRENGTH.

Btw the three new places where I added STRENGTH_NATURAL to the calculations were:
Lean BMI calculations in get_bmi_lean in character.cpp
The calculate hp code in recalc_hp in character.cpp
The part where strength affects nausea chance in modify_morale in consumption.cpp

As of making this fix, from what I could tell, weight_capacity in character.cpp uses get_str instead of get_str_base and thus does benefit from the STRENGTH enchantment.

I called the new enchantment STRENGTH_NATURAL based on my understanding of what it was supposed to represent. Feel free to suggest/change it to something else. A couple that came to my mind was STRENGTH_PERMANENT or STRENGTH_MUSCLE.

I saw that the base versus enchanted strength was used to in stat screens in such for whether the stats are colored green/red/yellow or such due to being buffed. If we'd rather that strength boosted from mutations is considered the natural state and then, say it being decreased by other factors should make it yellow/red even if it's at/above the character creation strength, then that would probably also involve updating the other 3 stats/mutations to behave similarly, is doable.

At first I considered adding STRENGTH_NATURAL as an alternative to STRENGTH (i.e. you don't need both), but it seemed less conventional to me to change activate_passive to modify strength with both enchantments in a way such that STRENGTH_NATURAL and STRENGTH stack like STRENGTH stacks with other STRENGTH enchantments, in the case that some mod or something used a multiplicative STRENGTH enchantment. Though I'd be willing to try.

It was a little unclear to me whether the Hyde trait's strength buff should count towards BMI/max hp as a physical enhancement, like whether it bulks them up physically, so I didn't add STRENGTH_PERMANENT to it yet.

#### Testing

1. Spawn a character with 10 strength.
2. Check the character's weight, like through a bathroom scale.
3. Check the character's max hp, like from character info.
4. Get the Insanely Strong mutation, bumping strength up to 17.
5. Check that the character's weight and max hp are now higher.
6. Verify you can now eat as much butter as you want until you're full, without getting nauseous, because the nausea chance of 8 is decreased by 17/2=8.

#### Additional context

Btw there were a couple of things that seemed odd to me about the nausea code, like how nausea chance was in the denominator so a lower nausea chance meant higher duration/intensity, or how the effect of strength on nausea chance jumps from -0% to -5% at 11, but I couldn't tell for certain whether those were intended or not.

Screenshots testing the fix:

Pre mutation:
Screenshot from 2026-06-30 19-57-07

image

Post mutation:

image

image

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -3962,7 +3962,12 @@
     "category": [ "SLIME" ],
     "enchantments": [
       {
-        "values": [ { "value": "MAX_HP", "add": 3 }, { "value": "STRENGTH", "add": -2 }, { "value": "DEXTERITY", "add": 3 } ]
+        "values": [
+          { "value": "MAX_HP", "add": 3 },
+          { "value": "STRENGTH", "add": -2 },
+          { "value": "STRENGTH_NATURAL", "add": -2 },
+          { "value": "DEXTERITY", "add": 3 }
+        ]
       }
     ]
   },
@@ -3984,7 +3989,12 @@
     "flags": [ "PSEUDOPOD_GRASP" ],
     "enchantments": [
       {
-        "values": [ { "value": "MAX_HP", "add": 6 }, { "value": "STRENGTH", "add": -4 }, { "value": "DEXTERITY", "add": 4 } ]
+        "values": [
+          { "value": "MAX_HP", "add": 6 },
+          { "value": "STRENGTH", "add": -4 },
+          { "value": "STRENGTH_NATURAL", "add": -4 },
+          { "value": "DEXTERITY", "add": 4 }
+        ]
       }
     ]
   },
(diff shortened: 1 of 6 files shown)


worm-girl:
Oh, this is unexpected, thank you! I tried once or twice to get this done but I kept getting stumped by how to make it work with the Alpha stat. This is a great solution.

PR #2745, opened 2026-07-15
Health fix

aattss:
#### Summary

Fix some health value logic seemingly being calculated incorrectly.

#### Purpose of change

So while I was looking into whether current hp% affected health regen yet in the character.cpp regen code, I noticed it does have an impact through its impact on health in the case that current hp is less than half of current hp, which seems like something that wouldn't happen while hp is positive.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/a4651547f7d4d6e1179cc5fb8ca1053ff48b3c27/src/character_body.cpp#L353-L358

I figured that while I was there, I'd read through the rest of the health code.

The toxin malus was positive the more toxins the player had, and was then maxed with a negative number, that's then added to health, so toxins wouldn't affect health.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/a4651547f7d4d6e1179cc5fb8ca1053ff48b3c27/src/character_body.cpp#L405-L414

Also, fatigue_regen_rate was cast from a float to an int even though it was a supposed to be a rate multiplier that seemed to range from 0.0 to 2.0.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/a4651547f7d4d6e1179cc5fb8ca1053ff48b3c27/src/character_body.cpp#L277-L281

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/a4651547f7d4d6e1179cc5fb8ca1053ff48b3c27/src/character.cpp#L5675-L5687

Also, if the player was not missing (i.e. did have) a low morale flag, then morale gets decreased, while otherwise, if they don't have the flag then it gets removed. It seemed to me it was probably intended that if they did have the flag then it should get removed, rather than if they have the flag then it doesn't get removed and just stays around.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/a4651547f7d4d6e1179cc5fb8ca1053ff48b3c27/src/character_body.cpp#L341-L350

#### Describe the solution

1. Change the wounded text from current hp < current hp / 2, to current hp < max hp / 2.
2. Change the toxin malus part to negative.
3. Change fatigue_regen_rate to a float, cast to - from to int (though I think this is currently called every turn), calculate roll_remainder( turns * fatigue_regen_rate ) ) so that 0.5 regen rate means that on average they get half of their turns to count rather than 0.5 rounding down to 0, and hten casting back to time_duration.
4. If they have the low morale flags, they lose health and also the flag, rather than the flag persisting. Unless their morale is low enough when that's calculated that they still meet the thresholds for low morale.

Btw in testing I noticed that in_sleep_state also includes when the player is trying to sleep, not just actual sleeping. Dunno whether that's supposed to count. Also I wasn't sure whether or not tmp_rates.recovery is allowed to be negative, like reverse sleep progress or something, though in the other place using .recovery for sleep stuff in character.cpp it seemed to have a >0 guard so I ended up going with that here too.

For morale, it made sense to me that if they still had low morale for the duration of the check, like when the new time period starts, the low morale flag remains, such as in the case where they have some long-lasting negative morale effect.

#### Testing

The player's health value can be seen with debug mode, though considering how many things affect it I just added logging for myself for the specific conditionals and values I was testing. The test is if low hp, high toxins, and a bunch of negative morale affect health, and if the health sleep calculations handle the fatigue recovery for decimal values rather than just truncating to 0/1. The hp/toxins/morale can be set in the debug menu.

### DIFF
--- a/src/character.h
+++ b/src/character.h
@@ -260,6 +260,15 @@ enum sleep_deprivation_levels {
     SLEEP_DEPRIVATION_MASSIVE = 14 * 24 * 60
 };
 
+/** @brief thresholds of morale at which health is impacted
+    @details If the player reaches low morale at some point in a day, their lifestyle score
+    is decreased.
+*/
+enum morale_levels {
+    MORALE_UNHEALTHY_LOW = -25,
+    MORALE_UNHEALTHY_VERY_LOW = -50
+};
+
 enum class blood_type {
     blood_O,
     blood_A,
--- a/src/character_body.cpp
+++ b/src/character_body.cpp
@@ -276,11 +276,13 @@ void Character::update_body( const time_point &from, const time_point &to )
         needs_rates tmp_rates;
         calc_sleep_recovery_rate( tmp_rates );
         const float fatigue_regen_rate = tmp_rates.recovery;
-        const int turns = to_turns<int>( to - from );
-        const time_duration effective_time_slept = time_duration::from_turns(
-                    roll_remainder( turns * fatigue_regen_rate ) );
-        mod_daily_sleep( effective_time_slept );
-        mod_continuous_sleep( effective_time_slept );
+        if( fatigue_regen_rate > 0.0f ) {
+            const int turns = to_turns<int>( to - from );
+            const time_duration effective_time_slept = time_duration::from_turns(
+                        roll_remainder( turns * fatigue_regen_rate ) );
+            mod_daily_sleep( effective_time_slept );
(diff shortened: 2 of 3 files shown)


worm-girl:
Thank you so much for this. I had to ditch automated tests because keeping up with them would have slowed development to a crawl, but you can see how that is a double edged sword here.

PR #2750, opened 2026-07-15
reduce inflatable decoration charges

andersonarp2810:
#### Summary
Found out inflatable dragon is not usable because it required more charges than its compatible battery could provide

#### Purpose of change
Makes inflatable dragon, inflatable ghost and fog machine usable

#### Describe the solution
Reduce charges_per_use on inflatable_decoration to 56. This also affects the inflatable ghost and fog machine, but they also use the same battery so it's not an issue

Maybe define the battery pocket on the inflatable_decoration to make sure future items who copy it work. 

#### Testing
Spawned an inflatable dragon, reloaded with lithium battery, was able to activate

#### Additional context
Before:
image

After:
image

### DIFF
--- a/data/json/items/items_holiday.json
+++ b/data/json/items/items_holiday.json
@@ -125,7 +125,7 @@
     "material": [ "plastic" ],
     "symbol": "P",
     "color": "light_red",
-    "charges_per_use": 100,
+    "charges_per_use": 56,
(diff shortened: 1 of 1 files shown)


andersonarp2810:
image
found a bug btw, activating these items do not destroy the item. one instance of inflatable ghost or dragon can spawn as many as you want as long as you have the batteries

PR #2756, opened 2026-07-17
Curses sound

JodiJodington:
#### Summary
Category "Brief description"
This allows you to build the curses interface with sound
#### Purpose of change
There is no reason that sound should be limited to when SDL is used as a gui. SDL has a very flexible subsystem system so why not use it.
#### Describe the solution
All this does is initialize the sound system on init_interface for ncurses instead of only for sdltiles and edits the CMakeLists.txt and Makefile to allow for SOUND=1 with TILE=0.
 
N/A

#### Testing
This compiles and works as expected on my linux machine. I don't have access to a windows or macos machine to do any testing on that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@
 #  make RELEASE=1
 # Tiles (uses SDL rather than ncurses)
 #  make TILES=1
-# Sound (requires SDL, so TILES must be enabled)
-#  make TILES=1 SOUND=1
+# Sound
+#  make SOUND=1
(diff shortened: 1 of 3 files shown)


JodiJodington:
it seems like the cmake compilation has some problems so I removed the part of this patch that touches CMakeLists.txt since that was blocking the CI. As far as I can tell, most of those linter warnings are about lines I never touched.

worm-girl:
Thanks!

PR #2761, opened 2026-07-18
Fix fire tiles appearing rotated in NeoDays tileset

PenwiseOnGit:
#### Summary

Single tile fires, the type you'd see in braziers, were still rendered as if they were rotated 90 degrees in NeoDays.  Presumably other fields have had a similar issue, but they were less blatant about it.

#### Purpose of change

A simple one-line fix to make single-tile fields draw properly in NeoDays.

#### Describe the solution

One of the render paths for drawing fields was missing its category (defaulting to NONE), which made the "rotates": false property on the tile get ignored here:

```cpp
if( !display_tile.rotates && category != TILE_CATEGORY::NONE &&
        category != TILE_CATEGORY::MONSTER )
```

This can also be fixed by giving the field tile an identical array of four sprites, but that'd need to be done for every field that doesn't rotate and seems to defeat the purpose of the rotates property.

#### Testing

Compiled with the change, made sure brazier fires were rendered properly in both NeoDays and MXotto+.  Did the same for a cloud of smoke around a smoker zombie to make sure it didn't somehow affect other fields rendering.


worm-girl:
thankyouthankyouthankyou. I did a ton of work on drawing code. It was some of the worst and messiest code in the repo (which is generally pretty good! DDA coding standards are high these days) and I tried fixing it up but I must have missed this when I was trying to condense it all down.

PR #2777, opened 2026-07-22
Fix possibly reading/writing invalid data while processing cigarettes.

PenwiseOnGit:
#### Summary

Hopefully fix an intermittent crash, or at least avoid unintended behaviour, while smoking cigarettes.

#### Purpose of change

Lit cigarette processing was removing itself in the middle of its processing and always returning false, which let the rest of item processing run on an object that's no longer valid.  That could be the reason some people are crashing while smoking.

#### Describe the solution

Just return true and let item processing handle the item's removal like it does for all other items.  Cigarettes didn't need to handle it themselves, and were unique in doing so.

Just leaving it.

#### Testing

Made a test character smoke a bunch of cigarettes and joints.  They all properly turned into dropped butts/roaches when finished.  I can't really confirm that this will fix the intermittent crashing, since I can't reproduce it, but it doesn't alter cigarette's behaviour and doesn't risk accessing invalid data.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -14111,6 +14111,7 @@ bool item::process_litcig( map &here, Character *carrier, const tripoint_bub_ms 
 {
     // cig dies out
     if( item_counter == 0 ) {
+        bool remove = false;
         if( carrier != nullptr ) {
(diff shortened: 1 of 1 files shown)


worm-girl:
Thanks!

PR #2778, opened 2026-07-22
Fix incorrect error_skipped_members error that happens sometimes

PenwiseOnGit:
#### Summary

Sometimes, especially when quitting and loading games repeatedly without closing the game, a json error will appear complaining about random json objects having unread keys despite being untrue and otherwise loading fine.  It's a harmless error, but annoying, and makes times when it's legitimate easier to miss.

#### Purpose of change

I tracked this down to a few potential issues in tiny_bitset's handling of its memory, along with how it was mapping index to bit locations.  In many cases, its data was left uninitialized, or left unchanged from resizing, which could easily mean a random bit was incorrectly flagged as true despite never actually being touched.  In its mapping function, whenever it had a block that was partially allocated for use, it was also excluding its last full-use block.  The two issues compounded to mean that, very randomly, a json object that was being loaded could have had an entry that was in a partial-use block, was populated with uninitialized or old data, and couldn't be properly written to.  When the error checker iterated over it, the bit wasn't set, so the error gets displayed.

#### Describe the solution

Fix up tiny_bitset's memory initialization to always set newly allocated/available bits to 0, and always start with zeroed data.  Fix tiny_bitset's copy constructor and operator to actually copy the entirety of its active blocks in the heap case.  Fix tiny_bitset's map function to not exclude the last block when it has partial blocks.  Fix tiny_bitset's resizing function not zeroing newly exposed bits.  Fix flexbuffer_json's method for reading the tiny_bitset to avoid potentially reading a block that is out-of-bounds.

Just leaving it.  I imagine most won't even encounter it, since I've only ever seen it when saving and loading games a bunch without exiting the game.  That's probably just developers and mod developers?

#### Testing

Spammed loading several different worlds with different mods, no json loading errors.  It's intermittent, so I can't guarantee it was actually fixed, but I also can't make it error with these fixes.  Edited a few json objects to include unread keys, and the unread keys did produce errors correctly, so these changes didn't break that functionality.

### DIFF
--- a/src/cata_bitset.cpp
+++ b/src/cata_bitset.cpp
@@ -16,28 +16,51 @@ void tiny_bitset::resize_heap( size_t requested_bits ) noexcept
 {
     size_t blocks_needed = bit_count_to_block_count( requested_bits );
 
+    const size_t old_size = size();
+
     // Round up to a multiple of the block allocation size.
     // The allocation size is stored by malloc so realloc() can be optimized if we keep resizing by small amounts.
+    // One extra block for prepending the capacity to the allocation, which is stored in the first block.
     blocks_needed = ( ( blocks_needed + ( kMinimumBlockPerAlloc - 1 ) ) / kMinimumBlockPerAlloc ) *
                     kMinimumBlockPerAlloc + 1;
 
+    block_t *data;
+
     if( is_inline() ) {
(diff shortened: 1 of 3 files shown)


worm-girl:
Thank you! This stuff is way over my head.

PR #2792, opened 2026-07-24
Add caution mode: toggleable warnings for fighting while impaired

msmfai:
I use this in my personal game but I believe it has wider utility as a way to guard against road hypnosis - which often gets people killed. The exact text and thresholds are a work in progress and this PR is mainly to gauge interest in adding this feature at all.

#### Summary
Features "Caution mode: optional warnings when you keep fighting while impaired"

#### Purpose of change

Players die by continuing to fight while slowed, hurt, or surrounded without noticing. Safe mode warns when a threat appears. Nothing warns when you keep going in a bad state.

#### Describe the solution

Adds a caution mode section to the options, next to safe mode. It is off by default. Each warning can be toggled on its own.

It warns you when:

- your speed is low and enemies are near
- you attack while slowed (asks for confirmation first)
- you take damage while slowed
- you get grabbed
- you are winded near enemies
- you are bleeding
- you have a bite or infection
- you are surrounded
- your pain is rising
- your torso or head is badly hurt
- you have been in danger a long time

Warnings are popups, so they stop you mid-keypress. Thresholds are configurable.

A JSON mod. It cannot add real options or stop an attack before it happens.

#### Testing

I played it and it works. Tuning the experience is pending affirmative interest in merging the concept.

### DIFF
--- a/src/avatar_action.cpp
+++ b/src/avatar_action.cpp
@@ -16,6 +16,7 @@
 #include "action.h"
 #include "activity_actor_definitions.h"
 #include "avatar.h"
+#include "caution.h"
 #include "bionics.h"
 #include "body_part_set.h"
 #include "calendar.h"
@@ -401,6 +402,9 @@ bool avatar_action::move( avatar &you, map &m, const tripoint_rel_ms &d )
                     return false;
                 }
             }
+            if( !caution::confirm_attack( you ) ) {
+                return false;
+            }
             you.melee_attack( critter, true );
             if( critter.is_hallucination() ) {
                 critter.die( &m, &you );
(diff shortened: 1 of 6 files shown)


msmfai:
Main motivation for this is that it's very easy to simply not notice you're chilly and wet and have 70 speed or that you're nearly out of stamina. In reality, pain forces you to pay attention to it, so I would also hope this adds some verisimilitude.

worm-girl:
Thanks! I tend to take a less is more approach and this is pretty maximalist, but the core idea is solid. I will go ahead and merge this and then pare out some of the stuff that isn't as necessary.

PR #2794, opened 2026-07-24
Add kiting mode: autoattack key strikes once then retreats

msmfai:
Kiting is simple but takes a lot of near-identical keypresses by hand. Two issues follow. The repeated input is a repetitive strain injury risk, and since kiting is precision rather than a decision, one mis-sequenced key makes you take damage for no real choice. This is an optional mode that repurposes the autoattack key (Tab by default) to do a whole kite step on one press, and warns you when the assumptions behind the kite no longer hold. It is off unless you turn it on, and turning it off returns Tab to normal autoattack. Behaviour is WIP, mostly posting to gauge interest.

#### Summary
Features "Kiting mode - an optional replacement for the autoattack key that performs one kite step per press"

#### Purpose of change

Kiting by hand is hundreds of repeated inputs. That is a repetitive strain injury risk, and because it is precision rather than judgement, one fumbled key costs HP for a non-decision.

#### Describe the solution

Adds a "Toggle kiting mode" keybinding. It is off by default. While on, the autoattack key (Tab by default) does one kite step per press instead of a normal autoattack.

- if you struck last press, retreat to the adjacent tile farthest from all enemies (orthogonal only, since diagonals cost more movement)
- otherwise, if an enemy is in weapon reach, strike once
- otherwise, hold and wait

If you are grabbed, knocked down, stunned, or winded, it shows a popup and does nothing. Cornered does the same. No game time passes on a popup. Hold the key to kite continuously, release to stop.

Toggle the mode off and the autoattack key behaves exactly as before.

- A self-driving kiting activity (activity_actor running until it stops itself). Built and scrapped, because it keeps its own state and keeps acting after the situation changes, so it could flail against a grab or walk into a fresh threat before an abort fired. The keypress version cannot, since the player is the loop.
- A JSON/EOC mod. It cannot add a keybinding or read movement-cost state, so it cannot express "retreat orthogonally when I struck last turn".

#### Testing

I played it and it works.

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -3014,6 +3014,12 @@
   },
   {
     "type": "keybinding",
+    "name": "Toggle kiting mode",
+    "category": "DEFAULTMODE",
+    "id": "toggle_autokite"
+  },
+  {
+    "type": "keybinding",
     "name": "Main menu",
     "category": "DEFAULTMODE",
     "id": "main_menu",
--- a/src/action.cpp
+++ b/src/action.cpp
@@ -290,6 +290,8 @@ std::string action_ident( action_id act )
             return "auto_travel_mode";
         case ACTION_TOGGLE_SAFEMODE:
             return "safemode";
+        case ACTION_TOGGLE_AUTOKITE:
+            return "toggle_autokite";
         case ACTION_TOGGLE_AUTOSAFE:
             return "autosafe";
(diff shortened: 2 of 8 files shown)


msmfai:
This is a hack I added for myself. If you like the concept and have comments I will clean the code up.

worm-girl:
I get the RSI issue here but I really don't want to automate any part of combat, ever, so this one's gonna be a no from me.

PR #2812, opened 2026-07-27
renamed social to socializing

JodiJodington:
#### Summary
Category "Brief description"
This edits skills.json to have the speech skill be called socializing instead of social for the purposes of learning the skill from the book. This might affect other things too but as far as I can tell, the only code that uses these names from the json is the book reading code.
#### Purpose of change
currently when reading a book that trains speech, the game says "You learn a bit about social!" which makes no sense.
"You learn a bit about socializing!" is still a bit weird but it makes grammatical sense.
#### Describe the solution
I edited skills.json,
I would have preferred "You learn a bit about being social" but since in the same file there's already a "tailor" -> "tailoring" transformation, I thought this would be appropriate.

#### Testing
I loaded the game with the new data directory, spawned a Car Buyer's Guide and the text printed as expected.

### DIFF
--- a/data/json/skills.json
+++ b/data/json/skills.json
@@ -69,7 +69,7 @@
   {
     "type": "skill",
     "id": "speech",
-    "name": { "str": "social" },
+    "name": { "str": "socializing" },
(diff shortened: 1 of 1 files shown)


worm-girl:
Makes sense to me!

PR #2836, opened 2026-07-30
remove the standalone zzip executable from the game

JodiJodington:
it can now be accessed via `cataclysm-tlg zzip` instead.

#### Summary
Category "Brief description"
Cataclysm comes with a tool called "zzip" which is not used internally. As far as I can tell, the only point of it being seperately installed is to allow people to decompress cataclysm saves without having the game installed, which isn't very sensical in my opinion.
#### Purpose of change
Instead of installing cataclysm-tlg *and* zzip, we now just install cataclysm-tlg which has a zzip entrypoint
running cataclysm-tlg opens cataclysm-tlg as normal, but running `cataclysm-tlg zzip` now redirects you to zzip's main() function
#### Describe the solution
We move tools/save/zzip_main.cpp into src/, rename main() to zzip_main(), and then do a quick check in the main.cpp main() that redirects you to zzip_main() if argv[1] is "zzip". 
I was originally just going to remove zzip entirely since cataclysm-tlg doesn't use it, but someone might want this for save editing purposes. The quality of the code and user experience is pretty attrocious and I can't imagine anyone actually uses zzip though, it doesn't even have a --help command, you can only learn how to use it by reading the main() function. I still think it'd be better to delete it personally.
#### Testing
I loaded a game, saved it, and loaded it, and all of it worked as expected.
#### Additional context

I was also going to expose a way to build it as a standalone executable with SEPERATE_ZZIP=1, but it turned out to be too annoying to modify the Makefile to allow for conditionally installing a binary so I didn't bother. If you want me to do that I can give it another try, I just assumed it wouldn't be something anyone would miss to be honest lol

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,15 +235,6 @@ ifneq (,$(findstring mingw32,$(CROSS)))
   JSON_FORMATTER_BIN=tools/format/json_formatter.exe
 endif
 
-# Determine zzip compression tool name
-ZZIP_BIN=zzip
-ifeq ($(MSYS2), 1)
-  ZZIP_BIN=zzip.exe
-endif
-ifneq (,$(findstring mingw32,$(CROSS)))
-  ZZIP_BIN=zzip.exe
-endif
-
 # Enable backtrace by default
 ifndef BACKTRACE
   # ...except not on native Windows builds, because the relevant headers are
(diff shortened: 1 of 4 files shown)


JodiJodington:
I wasn't able to try compiling it on windows which has a different main() definition. It should probably be fine, but if anyone's on windows and wants to try compiling this and say if it works, that would be great.

PR #2837, opened 2026-07-30
Fix a bug in my last PR that caused the game to report an error when trying to play more than two of the same sound at once

JodiJodington:
#### Summary
Category "Brief description"
There was a comment as evil as can be in sdlsound.cpp which mislead me into returning "true" instead of "false".
#### Purpose of change
Without this change, the debug.log would be filled with errors that weren't really errors. If built with BACKTRACE=1, this would also cause the game to freeze for a good amount of time to write out the backtrace.
#### Describe the solution
I corrected the comment and am now returning the correct value.
#### Testing
I loaded the game without my patch during a thunder storm and observed long pauses as well as errors in the debug.log due to lots of thunder sound effects, I then loaded the same game with my patch and observed no issues.


JodiJodington:
for the record, I checked git blame and the comment was just always wrong lol

worm-girl:
Thanks!

PR #2861, opened 2026-08-02
Made hiding/unhiding wires a bit nicer

JodiJodington:
#### Summary
Category "Brief description"
pressing `e` on a wall with revealed wiring will show an action to "hide/unhide wiring"
This PR makes the name dependent on if the wiring is hidden or not right now, this gives the bare minimum feedback for you to know that you actually did something by pressing the button (and it makes the name less clunky)
It now also mimics the behavior of "Reveal wall wiring" in the construction menu in that it asks the user if they want to hide or unhide all the wiring on the floor.
#### Purpose of change
Currently, you'd need to go over to each piece of revealed wiring and manually hide it or unhide it, which is very unfortunate if youre dealing with an entire floor where each tile on the wall is probably wired.
#### Describe the solution
the action is named "Hide wall wiring" if it is unhidden and "Unhide wall wiring" if it is hidden
to hide or unhide all wiring on the floor, we get a list of points in the Z level, iterate them and check if any of them are a vehicle that has a part with flag_WIRING, then set vehicle_part::hidden to the right value.
#### Testing
I loaded a new world, went into the evac shelter, revealed the wiring, and it all worked as expected.
#### Additional context
I also wanted to change the prompt in "Reveal wall wiring" from "Also reveal nearby wall wiring" to "Also reveal all wall wiring on this floor" since that's what it actually does, but I didn't want to undo the translation work people have done for something so minor.

### DIFF
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -541,11 +541,22 @@ void veh_app_interact::plug( map &here )
     }
 }
 
-void veh_app_interact::hide()
+void veh_app_interact::toggle_hide_wiring( map &here )
 {
     const int part_idx = veh->part_at( veh->coord_translate( a_point ) );
     vehicle_part &vp = veh->part( part_idx );
-    vp.hidden = !vp.hidden;
+    const bool should_hide = !vp.hidden;
+    vp.hidden = should_hide;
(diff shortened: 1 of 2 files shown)


worm-girl:
I have always found this whole system so ugly and clunky, stuff like this is a big help, thank you.

PR #2864, opened 2026-08-03
Making tracking dispersion affect dispersion

aattss:
#### Summary

Tracking dispersion didn't seem to be used in the gun firing code.

#### Purpose of change

I noticed that get_tracking_dispersion was called but not doing anything in fire_gun. As far as I can tell references aren't passed in.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/355e7e856376a39afa59d5b7d0d4011e7ae4546c/src/ranged.cpp#L1124-L1126

It seems that it was only used in a way that had an effect in calculate_ranged_changes, which was used for the UI/debug logging I think.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/355e7e856376a39afa59d5b7d0d4011e7ae4546c/src/ranged.cpp#L1919

#### Describe the solution

I put the return from get_tracking_disperesion into dispersion.add_range

            dispersion.add_range( get_tracking_dispersion( &gun, victim, true ).max() );

get_tracking_dispersion had rng logic within itself to modify tracking_dispersion, a double, and then did add_range( tracking_dispersion ) and returned that, so I figured it was trying to add the add_range( tracking_dispersion ) to the total dispersion and that the code in get_tracking_dispersion was itself correct.

#### Testing

Adding debug logging around the get_tracking_dispersion call in fire_gun might help? It's what I tested. Otherwise testing would probably involve a big gun against a close enemy.

### DIFF
--- a/src/ranged.cpp
+++ b/src/ranged.cpp
@@ -1122,7 +1122,7 @@ int Character::fire_gun( map &here, const tripoint_bub_ms &target, int shots, it
         // Add dispersion for shooting in close range.
         const Creature *victim = get_creature_tracker().creature_at( target, true );
         if( victim != nullptr ) {
-            get_tracking_dispersion( &gun, victim, true );
+            dispersion.add_range( get_tracking_dispersion( &gun, victim, true ).max() );
         }
         dealt_projectile_attack shot;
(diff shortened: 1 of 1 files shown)


worm-girl:
Whoops, yeah dispersion code is a headache. Thanks for both noticing and fixing that!

PR #2895, opened 2026-08-10
ru update 

Ritchie997:
@Манул made changes to the phone menu and settings.
And I still have 100 more words to go.


worm-girl:
Thanks!

PR #2989, opened 2026-08-24
Re-add rock throw to psycho killer and frenzied worker

aattss:
#### Summary

Let the pyscho killer and the worker/technician ferals throw rocks again (assuming it wasn't intended to take rock throwing away from those).

Feel free to disregard or close if that change was intended.

#### Purpose of change

I noticed in a playthrough that ferals weren't throwing any rocks at me. After some more investigation I found that some types of ferals do still through rocks, it's just that the ones name psycho killer and frenzied worker didn't, and I just happened to run into several of the psycho killer monster in a row.

Seems it was removed from those in #2924.

#### Describe the solution

I basically copied the rock throwing weapon from the monsters that still had it, back to mon_feral_human_knife and mon_feral_human_tool, who they were removed from.

I thought about if it would make sense for some ferals to have or not have rock throwing, but I was uncertain so I just stuck to creating a PR that adds it back to the ones it was removed from in #2924 just in case the removal was unintended.

#### Testing

Spawn in psycho killers and frenzied technicians, and verify that they throw rocks when in the correct range.

### DIFF
--- a/data/json/monsters/feral_humans.json
+++ b/data/json/monsters/feral_humans.json
@@ -155,6 +155,30 @@
     "description": "The Cataclysm began as an outbreak of what then seemed like violent insanity, but there must be something more to it, as these lost souls freely walk among the undead without being attacked.  This one has got their hands on a kitchen knife, and they look eager to put it to use.",
     "copy-from": "mon_feral_human_pipe",
     "special_attacks": [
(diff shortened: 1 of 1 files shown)


worm-girl:
These were removed on purpose.

PR #3009, opened 2026-08-26
Fix SDL audio callback crash

Anatidaephobic:
#### Summary
Bugfix "Fix SDL audio callback crash"

#### Purpose of change

A macOS crash report shows `EXC_BAD_ACCESS` on the `AudioQueue thread` inside `sound_effect_handler::slowed_time_effect`.

The callback checked whether time was slowed by reading the avatar's speed and moves. The main thread can change character state and its caches at the same time. That creates a data race and can crash the audio thread.

#### Describe the solution

Compute the time-slowed flag on the main thread around player actions and activities. Publish it through an atomic boolean. The audio callback now reads only that boolean.

This keeps the existing slowdown threshold and clears the flag when play ends.

Locking character state from the callback could block the audio thread and would require broader locking. Publishing one boolean keeps the callback independent of mutable game state.

#### Testing

- Built the macOS release target with tiles and sound enabled.
- Compiled `do_turn.cpp` and the no-sound stub path with warnings treated as errors.
- Ran `git diff --check`.

#### Additional context

Sanitized crash-log excerpt:

```text
Exception: EXC_BAD_ACCESS (SIGSEGV)
Crashed thread: AudioQueue thread

0  cataclysm-tlg-tiles                       + 15524920
1  cataclysm-tlg-tiles                       + 2760732
2  cataclysm-tlg-tiles                       + 2840188
3  sound_effect_handler::slowed_time_effect  + 88
4  Mix_DoEffects                             + 148
5  mix_channels                              + 564
6  SDL audio callback
```


worm-girl:
This one's been a real pain, thanks!

PR #3013, opened 2026-08-26
Fix nausea chance strength impact

aattss:
#### Summary
Fixing nausea chance strength reduction to be 0-10 for strength 10-19.

#### Purpose of change
Previously, at strength 10, nausea chance was not reduced, and at strength 11, it was. And it was reduced by about strength/2=5.

This fix attempt to implement the comment at https://github.com/Cataclysm-TLG/Cataclysm-TLG/pull/2990#issuecomment-5403228688

#### Describe the solution
Now, at strength 9, nausea chance is still not reduced, at strength 10 it's reduced by 1, and each additional point in strength reduces nausea chance by an additional point, up to reducing nausea chance by 10 at strength 19.

Capping at strength 20 instead? There are other possibilities for the range, like starting at 11 instead of 10, or having strength below 9 increase nausea chance for nausea-inducing food. This is just attempting to implement the behavior as described in the comment.

#### Testing
Test that a strength 10 character does not have significantly less nausea chance/duration than a strength 9 character, and that a strength 20 character is decently less likely to get nausea or gets not as severe nausea.

### DIFF
--- a/src/consumption.cpp
+++ b/src/consumption.cpp
@@ -1490,9 +1490,9 @@ void Character::modify_morale( item &food, const int nutr )
         }
         int strength_adjusted = enchantment_cache->modify_value( enchant_vals::mod::STRENGTH_NATURAL,
                                 get_str_base() );
-        if( strength_adjusted > 10 ) {
-            nausea_chance -= ( std::min( get_str(), strength_adjusted ) / 2 );
-        }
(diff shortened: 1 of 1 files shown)


worm-girl:
Thank you! And it's meant to be a pretty minor thing, it's probably fine that it caps at 19 and not 20.

PR #3047, opened 2026-09-03
Remove "MINEABLE" from most terrain, remove auto-mine.

worm-girl:
#### Summary
Remove "MINEABLE" from most terrain, remove auto-mine.

#### Purpose of change
- Despite making jackhammers unable to mine through walls, the humble pickaxe was still able to bore through pretty much everything without any risk of collapse or any other ill effects. It did so extremely quickly, with basically no drawbacks or risks. That is not how any of this works.

#### Describe the solution
- Most terrain and furniture is no longer mineable. Exceptions include the stone slab, the three sizes of boulder, and pavement, outdoors, not on bridges.
- Removed auto-mining. This was an ancient and very silly feature from back when you could just walk through walls basically.
- Greatly buffed the bash damage on both pickaxes, and reduced the stab damage. It's reasonable that a pickaxe could smash through a wall, certainly, which is what the smash command is for. Now they're better at doing that.
- Fixed the bronze pickaxe's to-hit and weight, it is now more like its steel counterpart.
- Added DIG 1 to both pickaxes, allowing you to use them for basic digging tasks.
- Currently this means solid earth and solid rock can no longer be dug out except by explosives. I don't think I have a problem with this. Digging solid bedrock is effectively impossible, and digging any kind of a cellar in solid earth would quickly cause cave-ins without properly installed supports. That means it'd have to be a multi-step construction rather than just 30 minutes per tile of you magically boring the place out.


Sh1noskay:
This is not a protest, I dont have an actual opinion on removing mineable. 

Just had to comment on the "almost impossible" part. 

As a previous cable dog in the military, I have spent quite a lot of time pickaxing through packed earth and gravel. I can get probably 3 to 6 feet in an hour or three.... (Though more times than one, I did it in 30 minutes) depending on how good or bad I'm feeling that day. I was a bit of an exception, granted (a beast with a hammer and a pickaxe) but it's not at all impossible. 

bashing through does make sense though... "Mountain Man" of India used a hammer to bore through... same with John Henry.

PR #3055, opened 2026-09-03
duffelbag size fixes

dynst:
#### Summary
Content "Fix survivor duffel bag, delete 'longarm bag'"

#### Purpose of change

A `duffelbag`'s size was corrected to be larger than 35L in #1796 but the crafted item, the survivor duffel bag, wasn't updated.

The craftable-only "longarm bag" item does not exist in the real world. The term "longarm bag" did not exist until a dubious PR for DDA invented it for the `long_duffelbag`.

#### Describe the solution

Set survivor duffel to 10% larger than the baseline duffelbag.

Delete recipe and replace every `long_duffelbag` with a `duffelbag`.

#### Additional context

Item was obsoleted in DDA in https://github.com/CleverRaven/Cataclysm-DDA/pull/81883

### DIFF
--- a/data/json/items/armor/bespoke_armor/custom_storage.json
+++ b/data/json/items/armor/bespoke_armor/custom_storage.json
@@ -16,7 +16,7 @@
     "pocket_data": [
       {
         "pocket_type": "CONTAINER",
-        "max_contains_volume": "38 L",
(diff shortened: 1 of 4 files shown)


worm-girl:
This item is slated for removal and no change is needed.

PR #3056, opened 2026-09-03
delete extra copy of trash_junkyard itemgroup

dynst:
#### Summary
None

#### Purpose of change

When the file `SUS/trash.json` was manually added to make the backport of d6f58f1a24415c035dbbcbe91b54380621459205 work in #1468, it included the `trash_junkyard` itemgroup from DDA.

But that itemgroup was deleted in DDA from this older file in https://github.com/CleverRaven/Cataclysm-DDA/pull/79855 and moved into the new file in https://github.com/CleverRaven/Cataclysm-DDA/pull/79848.

https://github.com/Cataclysm-TLG/Cataclysm-TLG/blob/f4bbd87aa5f0408214b00a91fa519738de347dee/data/json/itemgroups/SUS/trash.json#L122-L123

TLG includes the new file, but never backported the deletion, so it now has 2 itemgroups with the same name in different files.

#### Describe the solution

Delete the older copy.

Delete the newer copy? (Edit: done in 7d2bb903127f0886d072939ad11d068087a0c5f1)

#### Additional context

It's unclear why 2 itemgroups with the same name didn't trigger any kind of error.

### DIFF
--- a/data/json/itemgroups/trash_and_debris.json
+++ b/data/json/itemgroups/trash_and_debris.json
@@ -303,26 +303,6 @@
     ]
   },
   {
-    "type": "item_group",
-    "id": "trash_junkyard",
-    "items": [
-      { "item": "glass_shard", "prob": 30, "charges": [ 1, 60 ] },
-      [ "scrap", 50 ],
-      [ "steel_chunk", 30 ],
-      [ "steel_fragment", 30 ],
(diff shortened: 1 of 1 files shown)


worm-girl:
Good catch, but you deleted the wrong one. I've deleted the other one instead.

PR #3062, opened 2026-09-05
Extract missing JSON translation fields

eugene-lesnov:
#### Summary
Bugfixes "Extract missing translatable strings from JSON"

#### Purpose of change
While working on the Russian .po file and checking the translation in-game, I noticed that some strings remained in English. These strings come from JSON data but were not picked up by the translation extractor.

#### Describe the solution
Added extraction for the following JSON fields:

- blood_analysis_description
- acid_dmg_verb
- cant_remove_reason
- radiation_hurt_message
- unattend_message in recipe steps

None
#### Testing
Ran `lang/update_pot.sh` locally. The updated extractor produced 74 new translation entries and removed none.

#### Additional context
Translation files are not included because I am still checking for other strings that may be missing from the extraction process.
This is a large project, and I am still getting familiar with some parts of its translation system.
Sorry if I missed any project conventions or if something needs to be adjusted.

### DIFF
--- a/lang/string_extractor/parsers/bionic.py
+++ b/lang/string_extractor/parsers/bionic.py
@@ -11,3 +11,8 @@ def parse_bionic(json, origin):
     if "description" in json:
         write_text(json["description"], origin, c_format=False,
                    comment="Description of bionic \"{}\"".format(name))
+
+    if "cant_remove_reason" in json:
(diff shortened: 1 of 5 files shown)


worm-girl:
edit: d'oh, nevermind what I wrote before. Thank you so much!

PR #3072, opened 2026-09-07
Localize messages

eugene-lesnov:
#### Summary
I’m not fully confident about these changes, so I’d really appreciate a thorough review. 
My background is in a different technology stack..But I hope the approaches are clear enough.

#### Purpose of change
Some strings were not available for translation.

#### Describe the solution
- Translate the Medical menu title.
- Translate the complete electronic-port message.
- Extract values from `e_port` and `e_ports_banned` with one
  `electronic port type` context, then translate them at display time.
- Translate port names only when displaying them; keep the IDs used by game logic unchanged.
- Extract JSON flag display names for translation.

#### Testing
- Compiled with -Werror flag.
- Ran the `lang/update_pot.sh` (locally) pipeline for JSON flag names.

### DIFF
--- a/lang/string_extractor/parsers/generic.py
+++ b/lang/string_extractor/parsers/generic.py
@@ -32,6 +32,14 @@ def parse_generic(json, origin):
         write_text(json["description"], origin, c_format=False,
                    comment=comment + ["Description of \"{}\"".format(name)])
 
+    port_comment = "Electronic port type of item \"{}\"".format(name)
+    if "e_port" in json:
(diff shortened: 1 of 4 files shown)


worm-girl:
The C++ looks good. I'm not actually all that familiar with python beyond kicking the scripts until they work, and they look fine to me. I'll merge this and then run the string extractor and see if the game builds a new release. That'll tell us whether there were any problems or not.

edit: It worked! So I guess this was all good. Thank you!

PR #3077, opened 2026-09-08
Localize messages

eugene-lesnov:
#### Summary

#### Purpose of change
Some strings were not available for translation.

#### Describe the solution
Each commit fixes one localization problem:

- The companion return prompt is now extracted where it is defined instead of relying on a duplicate elsewhere.
- The full warning about unplaced overmap specials can now be translated without including the generated list.
- The missing mapgen update error is now included in the translation catalog.
- The overmap route hint is translated before the current key is inserted.
- Fault suffixes such as `snapped` and `no handle` are now extracted from JSON.
- Faction names, including `Your Followers`, are now translated in epilogue titles.

#### Testing
- Compiled the affected C++ files with `-Werror`
- Ran the full POT generation pipeline, including JSON extraction and catalog validation.

### DIFF
--- a/lang/string_extractor/parsers/fault.py
+++ b/lang/string_extractor/parsers/fault.py
@@ -10,3 +10,6 @@ def parse_fault(json, origin):
     if "item_prefix" in json:
         write_text(json["item_prefix"], origin,
                    comment="Prefix for item's name \"{}\"".format(name))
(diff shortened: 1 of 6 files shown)


worm-girl:
Thank you!

PR #3089, opened 2026-09-12
Chinese translation completed

CN-Xuan:
#### Summary
Add Chinese translations

#### Describe 
-本次新增词条已全部翻译完成，主要涉及 lang/po/zh_CN.po
-由于目前仅有我一人负责 CTLG 的中文翻译，能力和时间有限，不排除部分翻译描述存在不准确的问题
-后续会在游玩过程中边玩边修正不准确的翻译

-All newly added entries have been translated, mainly affecting lang/po/zh_CN.po
-As I am currently the only one working on the Chinese translation for CTLG, with limited time and capacity, some translations may be inaccurate
-I will continue to revise any inaccurate translations while playing in the future.

### DIFF
--- a/lang/po/zh_CN.po
+++ b/lang/po/zh_CN.po
@@ -117,7 +117,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: cataclysm-tlg 1.0\n"
 "POT-Creation-Date: 2026-09-07 18:55-0700\n"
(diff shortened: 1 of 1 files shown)


worm-girl:
Thank you so much! I ran the extraction script right after this in #3091 so there may be a few more that can be filled in now. See #3077 if you're curious which ones.

PR #3094, opened 2026-09-13
Advance car cabin heating.

Maepple:
#### Summary
Category "Brief description"

Separates electric heaters and coolant heater cores into separate parts and functions.

#### Purpose of change

Car heating repair are relatively routine and simple thing to fix, so doesn't make much sense for them to be no install.

Looking into it, I realised I could further develop the way the system works. The mechanics of most car's heaters are through pushing heat from the coolant through a smaller radiator, that has a fan to blow in, however, there are hybrid cars and electric cars that do not have heater cores to push heat through the system, and have electric heaters.

#### Describe the solution
This change adds an electric integrated heater, and I added functionality to the existing car heater so it can only run when the car has a combustion engine, that is active, and is currently running to function as a heater, It simulates the inability to get heat out of the system when it's not running as the water pump that pushes coolant around is no longer running.

Electric cars have a replaced electric heater that functions the same as the old system. Hybrids also have the electric heater system.

Leaving the system as it is?

#### Testing

Compiled game on Linux with make, looked around at cars finding electric ones and combustion engine cars, combustion engine cars only output heat when the heater is activated when the engine is running and is a combustion engine, hybrids with new heating only function when the combustion engine is running, even when electric functions.

#### Additional context
Not sure if you want this Worm but I found this fun to create.


worm-girl:
That's not how car heaters work IRL. There's a reason they're implemented as they are currently.

Maepple:
> That's not how car heaters work IRL. There's a reason they're implemented as they are currently.

If this is not how car heaters work IRL, please enlighten me.

[Look up how car heater cores function](https://www.cars.com/auto-repair/glossary/heater-core/), they're hooked up to a thermostat that opens a valve that pushes coolant through another smaller radiator(This is a heater core, sometimes called a heater matrix), which is close to the car cabin, which has a fan that blows air through it into the cabin. Thus coolant that is being used to keep engines at the proper temperature, goes into heating a car.

Replacing the coolant in a car, you have to heat up the coolant, with the heat setting cranked up in the car, which opens the valve that has coolant in the heater core loop that you want to purge.

Basically no cars outside of electric and hybrids have electric heaters in them.
Cars made in 2024 still have these radiators for coolant to heat up the cabin.

worm-girl:
? Nobody said anything about electric heaters. It's just a fan that blows the hot air off of the radiator and into the cabin, which is what already exists in the game. That's why you can't take the integrated heater out of the engine and use it elsewhere. Your link over-explains it by trying to explain the entire radiator system which made your in-game descriptions confusing.

Anyway this is the sort of thing you really ought to ping me over before you just run off and contribute, I'm sorry to close this after you've already done all the work, but I would have told you not to do it to begin with.

servant-of-the-chihuahua:
> ? Nobody said anything about electric heaters. It's just a fan that blows the hot air off of the radiator and into the cabin, which is what already exists in the game. That's why you can't take the integrated heater out of the engine and use it elsewhere. Your link over-explains it by trying to explain the entire radiator system which made your in-game descriptions confusing.
> 
> Anyway this is the sort of thing you really ought to ping me over before you just run off and contribute, I'm sorry to close this after you've already done all the work, but I would have told you not to do it to begin with.

Just wanted to make a comment on this because I think it is relevant to survival gameplay. To be clear my experience comes from cdda, I was just looking around here and eyeballing the commits to see how this fork differs. It is something I have been meaning to give a try! So there may be changes in this fork I am unaware of.

What op said about how heaters work is generally accurate. You have the waste heat of the engine being captured by the coolant which is then pushed by a water pump into the heater core and is bled into the cabin. So without an engine that is running, you will run out of heat because the coolant is not being heated or circulated by the engine. The same is basically true for the AC because it needs the compressor to be running. Electric/Hybrid vehicles do not tend to have that restriction which is one of their advantages.

This is relevant to survival gameplay because fuel is a much more precious resource than electrical power. You can just slap a solar panel down and largely beat the climate currently. If you have to burn fuel for that heating to occur then it changes the dynamics a bit and makes it less of a no brainer and adds some additional challenge to the survival aspects of the game. So I suppose I can see why one might want a running engine to be a requirement for IC Engines. 

I can also see why overcomplicating it might be undesirable. Expanding out the system a bit does add more items and considerations such as electric heaters that you can plug into ICE cars and the like. It could also be argued that adding that restriction would largely only affect the early game and would be played around in the late game and isn't worth it. I just wanted to toss my two cents in because I can see the appeal from a survival standpoint for wanting ICE to require the engine to be running for the climate functions to work.

worm-girl:
The game already has heaters that work like how car heaters work irl.

servant-of-the-chihuahua:
> The game already has heaters that work like how car heaters work irl.

Hmm, I feel like we might be misunderstanding eachother a bit here. Could you clarify what you mean when you say they work the same way that car heaters work irl? Is something different with how they work in this fork vs cdda? In cdda you pretty much turn on the heater and it drains car battery and provides heat even when the engine is off.

In real life the heater in a ICE car needs hot coolant to provide the heat. So with the engine off the coolant you are extracting the heat from would run out of thermal energy pretty quickly and the heater would pretty much just be a fan at that point.

I certainly don't intend to annoy or try to pressure you into the change. I just figured it could be a fun addition to survival and now I kinda feel like we aren't on exactly the same page when it comes to how heaters work. I suspect there may be something about how heaters work in TLG that I am unaware of.

worm-girl:
This pr is closed.

PR #3112, opened 2026-09-20
Localization work

eugene-lesnov:
#### Summary

Fixes several messages that were difficult or impossible to translate correctly.

#### Purpose of change

Some messages had broken placeholders, untranslated fragments, internal IDs, or English-specific grammar.

#### Describe the solution

- Fix two Megafauna messages with an invalid `%`.
- Keep one `%s`, matching the actual argument count.
- Replace untranslated field message fragments with complete strings.
- Remove the internal `ACT_*` value and fix the player/NPC message arguments.
- Remove possessive placeholders from the bionic eye message.

The messages could have kept using translated sentence fragments and possessive runtime values, but translators would still need to work around unknown grammar and word order. Complete messages are simpler and safer to translate.

#### Testing

- Compiled both changed C++ files with localization and `-Werror`.
- Validated both JSON files.

#### Additional context

I’m so glad to be able to contribute to the development.
I don’t want to break anything, please let me know if I missed something or should change anything.


data/json/bionics.json
```diff
@@ -420,7 +420,7 @@
     "description": "The eyes have been entirely replaced by a pair of implanted cameras which provide clear, high-resolution vision and come with a built-in zoom that allows the user to see long distances and treat all ranged weapons as though they had a scope.  If unpowered, the user will be completely blind.",
     "occupied_bodyparts": [ [ "eyes", 2 ] ],
     "flags": [ "ENHANCED_VISION", "BIONIC_TOGGLED", "BIONIC_SLEEP_FRIENDLY" ],
-    "cant_remove_reason": "The CBM has completely replaced %1$s eyes now.  Removing them would leave %2$s blind.",
+    "cant_remove_reason": "The CBM has completely replaced the patient's eyes.  Removing it would leave the patient blind.",
```
eugene-lesnov, line 423: As I’m not a native English speaker, I’m not completely sure which wording sounds more natural..

Maybe it would be better to use: “The CBM has completely replaced the eyes. Removing it would cause permanent blindness.”

data/mods/Megafauna/mf_weakpoint_sets.json
```diff
@@ -87,14 +87,14 @@
             "effect": "nausea",
             "permanent": true,
             "chance": 10,
-            "message": "You have ripped the %s bowels!  The % is retching.",
+            "message": "The %s starts retching after its bowels are ruptured.",
```
eugene-lesnov, line 90: The original message had a broken % placeholder and only received one argument. I rewrote it to use one valid %s and keep the same meaning as closely as I could.

data/mods/Megafauna/mf_weakpoint_sets.json
```diff
@@ -87,14 +87,14 @@
...
             "damage_required": [ 5, 25 ]
           },
           {
             "effect": "winded",
             "permanent": true,
             "chance": 10,
-            "message": "You have punctured the %s lungs!  The % has trouble breathing.",
+            "message": "The %s has trouble breathing after its lungs are punctured.",
```
eugene-lesnov, line 97: This message had the same problem with a broken % placeholder.

